Issue #12, opened 2020-12-23
Add commandline tool example


OTL:
works on urdf-viz/ROS on win/mac/linux

mitsuharu-kojima:
Remaining Tasks
- [ ] Add samples( cf https://github.com/openrr/openrr/tree/main/openrr-planner )
  - [ ] PR2
  - [ ] UR10
- [ ] Factor out robot_command

PR #41, opened 2021-01-05
[openrr-client] Add configs to create IkClient

mitsuharu-kojima:
refs #12

### DIFF
--- a/openrr-client/src/clients/ik_client.rs
+++ b/openrr-client/src/clients/ik_client.rs
@@ -2,7 +2,11 @@ use arci::{Error, JointTrajectoryClient, TrajectoryPoint};
 use async_trait::async_trait;
 use k::nalgebra as na;
 use k::Isometry3;
-use std::sync::Arc;
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, sync::Arc};
+use tokio::sync::Mutex;
+
+type ArcJointTrajectoryClient = Arc<dyn JointTrajectoryClient>;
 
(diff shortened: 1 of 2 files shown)


openrr-client/src/clients/ik_client.rs
```diff
@@ -235,3 +239,87 @@ pub fn check_ik_with_interpolation_and_constraints(
...
+            Arc::new(create_jacobian_ik_solver(&parameters))
+        },
+    )
+}
+
+pub fn create_ik_client(
+    client: ArcJointTrajectoryClient,
+    full_chain: &k::Chain<f64>,
+    config: &IkSolverConfig,
+) -> IkClient<ArcJointTrajectoryClient> {
```
OTL, line 295: Comment:
It might be better to return `Result<IkClient<>>` to remove `unwrap()` someday.
The function name will become `fn try_create_ik_client`.
Or add new constructor for `IkClient::try_new_wtih_config()`.

openrr-client/src/clients/ik_client.rs
```diff
@@ -235,3 +239,87 @@ pub fn check_ik_with_interpolation_and_constraints(
...
+    pub allowable_position_error_m: f64,
+    pub allowable_angle_error_rad: f64,
+    pub jacobian_multiplier: f64,
+    pub num_max_try: usize,
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct IkClientConfig {
+    pub name: String,
+    pub client_name: String,
```
OTL, line 257: Comment:
As same as before PR, name sounds a little ambiguous for me.
I don't understand whole structure actually, I can't say that what is the best.
(name -> xxx_key ?)

OTL (approved):

PR #43, opened 2021-01-06
Update tokio to 1.0

### LINKED ISSUE #19: Use tokio=1.0
tokio is already reached version 1.0.0.
We should use it instead of 0.2 or 0.3.

taiki-e:
Closes #19

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
 
 jobs:
   windows:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,3 +27,6 @@ members = [
 "openrr-planner" = {path = "openrr-planner"}
 "openrr-sleep" = {path = "openrr-sleep"}
 "openrr-teleop" = {path = "openrr-teleop"}
+
+# https://github.com/sebcrozet/kiss3d/issues/262
+kiss3d = { git = "https://github.com/sebcrozet/kiss3d", rev = "6274d50" }
--- a/arci-gamepad-gilrs/Cargo.toml
+++ b/arci-gamepad-gilrs/Cargo.toml
@@ -11,4 +11,4 @@ crossbeam-channel = "0.5.0"
 log = "0.4"
 
 [dev-dependencies]
-tokio = { version = "0.3", features = ["full"] }
+tokio = { version = "1.0", features = ["full"] }
--- a/arci-ros/Cargo.toml
+++ b/arci-ros/Cargo.toml
@@ -14,7 +14,7 @@ anyhow = "1.0"
 paste = "1.0"
 async-trait = "0.1"
 crossbeam-channel = "0.5.0"
-tokio = { version = "0.3", features = ["full"] }
+tokio = { version = "1.0", features = ["full"] }
 serde = { version = "1", features = ["derive"] }
 
 [dev-dependencies]
--- a/arci-urdf-viz/Cargo.toml
+++ b/arci-urdf-viz/Cargo.toml
@@ -6,13 +6,15 @@ edition = "2018"
 
 [dependencies]
 arci = { version = "0.0.1", path = "../arci" }
-reqwest = { version = "0.11", features = ["blocking", "json"] }
+ureq = { version = "2", features = ["json"] }
 serde = { version = "1.0", features = ["derive"] }
 anyhow = "1.0"
 openrr-planner = {version = "0.0.1", default-features=false}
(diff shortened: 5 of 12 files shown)


OTL:
Thank you for great contribution!
Changing the interface affects a lot for the code out side of this repo, so it is difficult to merge soon.

taiki-e:
> Changing the interface affects a lot for the code out side of this repo, so it is difficult to merge soon.

Indeed. It's a bit inefficient, but I can also remove API changes other than the version change of the runtime (by using another way mentioned in the second commit).

(However, in any case, if these libraries have code that depends on the executor or reactor of tokio 0.3, the version change of the runtime is a breaking change.)

OTL:
I read the comment of the seccond commit, and I understand what you mean.

The status of openrr is far from your expectation, the problem of runtime sounds smaller than the functionality for me. (Threre are much more problems actually.)
One of the most important problem is the interface of `arci::JointTrajectoryClient` which you are changing.
I think that it may be OK to change it, but we have to discuss about it in the view of robot control system.
(It is basicly possitive to change into async.)

And we lack the description about the trait!! (It is also one of the most important things))

taiki-e:
@OTL Thank you for the explanation. On second thought, this is a very aggressive change, and probably it won't bring a performance improvement worth its cost.

So I'll consider a way that doesn't cause breakage...

First, the problem I am currently aware of is:

- The current problem is that the use of reqwest blocking API (`reqwest::blocking`) interferences with the use of tokio on the top-level. [ref1](https://github.com/seanmonstar/reqwest/blob/bd9ff9f371b756decf26fbbde1687433b0f63774/src/blocking/wait.rs#L67), [ref2](https://users.rust-lang.org/t/how-to-make-a-blocking-call-from-a-warp-function/50099/4)
  (Currently, this should only happen if debug assertion is enabled, but I'm not sure if it's guaranteed not to happen if debug assertion is disabled.)

AFAIK, here are the main ways to avoid this:

- Call `reqwest::blocking` in a separate thread.
  - The problem of this way isn't the only performance, we need to be careful of this issue on every call of `reqwest::blocking`, and its misuse is not well detected by testing.
    (When I first wrote this patch, there were actually about 4 `reqwest::blocking` uses, but the test detected only one.)
    Test coverage is another issue, but that's another reason I chose the second method first. (I fagot to explain it first, sorry.)
- Use `reqwest` instead of `reqwest::blocking` in async context.
  - This way can fix the root problem and it is more efficient usually than the first way, but this way needs aggressive change and is controversial. And probably this won't bring a performance improvement worth its cost.
- Use an HTTP client other than reqwest. (e.g., [curl-rust](https://github.com/alexcrichton/curl-rust), [ureq](https://github.com/algesten/ureq), see also [this post by Michiel Mulders](https://blog.logrocket.com/the-state-of-rust-http-clients/))
  - This way also can fix the root problem.

Third way is an option I missed when I first wrote this patch.
And that's what I currently think is the preferred solution.

What do you think? (I can create patches for all versions if needed.)

taiki-e:
> - Call `reqwest::blocking` in a separate thread.

By the way, here is an implementation example of this: https://github.com/taiki-e/openrr/commit/7ebd7c1a36f991ba19a743c946ecd62d5e81f301

taiki-e:
> - Use an HTTP client other than reqwest. (e.g., [curl-rust](https://github.com/alexcrichton/curl-rust), [ureq](https://github.com/algesten/ureq), see also [this post by Michiel Mulders](https://blog.logrocket.com/the-state-of-rust-http-clients/))

And, here is an implementation example of this (using ureq): https://github.com/taiki-e/openrr/commit/aabfcdfe191a0ea762703406d980b52a3725446b

([I needed to use `serde_json::to_value`](https://github.com/taiki-e/openrr/commit/aabfcdfe191a0ea762703406d980b52a3725446b#diff-64bda139809d4a4d4cb3ba44ee7bd2d28714cf97bdb03298b05aea52d11b4409R48) because it doesn't seem to be able to send arbitrary json. Also, note that both curl-rust and ureq take `&str` as an URL instead of `url::Url`.)

OTL:
I preffer ureq implementation. Thanks a lot!!

(In the future, we might migrate to `async current_joint_positions` and use `reqwest::Client`, but the purpose of urdf-viz-web is minimum, simple http client.)

taiki-e:
Updated to use ureq. (71eebdc)

Also, I adjusted the test a bit (7f85eaf) because 71eebdc caused [a connection error in the test](https://github.com/openrr/openrr/runs/1706648881?check_suite_focus=true) (probably due to ureq sending the request earlier than the web server started).

OTL (approved):
Thanks a lot for your patience and contribution!

PR #54, opened 2021-01-13
[arci-ros] Support anonymous init

mitsuharu-kojima:


arci-ros/src/rosrust_utils.rs
```diff
@@ -255,3 +256,13 @@ macro_rules! define_action_client {
         );
     };
 }
+
+pub fn init(name: &str, anonymous: bool) {
```
OTL, line 260: I want to keep it simple `arci_ros::init()`.
This should be named as `init_with_anonymized()` without a bool argument.

When you want to use bool to switch, I think below is acceptable basicly.

```rust
if is_anon {
  arci_ros::init_with_anonymized(&name);
} else {
  arci_ros::init(&name);
}
```

PR #55, opened 2021-01-13
[openrr-apps] Support speak in robot_command

mitsuharu-kojima:


openrr-apps/src/robot_command.rs
```diff
@@ -81,6 +81,8 @@ pub enum RobotSubCommand {
     },
     /// List available clients.
     List {},
+    /// Speak.
```
OTL, line 84:
```rust
/// Speak text message
```

openrr-apps/src/robot_config.rs
```diff
@@ -17,6 +17,9 @@ pub struct RobotConfig {
     pub ros_clients_configs: Vec<RosControlClientConfig>,
     pub urdf_viz_clients_configs: Vec<UrdfVizWebClientConfig>,
 
+    #[cfg(feature = "ros")]
+    pub ros_espeak_client_config: Option<RosEspeakClientConfig>,
+
```
OTL, line 22: I think that it is OK to exist ros_***_config even if ros feature is off. (If it does not cause compile error or some problems)

OTL (approved):

openrr-apps/src/robot_config.rs
mitsuharu-kojima: RosEspeakClientConfig is defined in arci-ros.
So I thinks we cloud not compile without feature=ros

PR #72, opened 2021-01-21
[arci, arci-ros, arci-urdf-viz] Add frame_id as send_pose arg in Navi…

mitsuharu-kojima:
…gation trait


OTL (approved):

arci-ros/src/ros_nav_client.rs
```diff
@@ -179,6 +177,7 @@ impl Navigation for RosNavClient {
     async fn send_pose(
```
OTL, line 177: I'll rename this to `move_to` or something....

PR #74, opened 2021-01-21
[arci-ros] Add RosNavClientConfig

mitsuharu-kojima:


arci-ros/src/ros_nav_client.rs
```diff
@@ -137,6 +138,11 @@ impl RosNavClient {
             nomotion_update_client,
         }
     }
+    pub fn new_from_config(config: RosNavClientConfig) -> Self {
+        let mut s = Self::new(config.request_final_nomotion_update_hack);
+        s.clear_costmap_before_start = config.clear_costmap_before_start;
```
OTL, line 143: Comment:

As you know this should be an argument as same as request_final_nomotion_update_hack

OTL (approved):

PR #77, opened 2021-01-21
[openrr-apps] Supprot Navigation / MoveBase in robot_command

mitsuharu-kojima:


openrr-apps/src/robot_command.rs
```diff
@@ -98,6 +101,28 @@ pub enum RobotSubCommand {
     Speak { message: Vec<String> },
     /// Execute an external command.
     ExecuteCommand { command: Vec<String> },
+    /// Get navigation current pose.
+    GetNavigationCurrentPose {},
```
OTL, line 105: Is it possible to `GetNavigationCurrentPose {}` -> `GetNavigationCurrentPose`?

It is ok to do it in another change for all members.

OTL (commented):
Please fix commit message typo at frist

openrr-apps/src/robot_client.rs
```diff
@@ -95,6 +97,28 @@ impl RobotClient {
...
+                Some(Arc::new(RosNavClient::new_from_config(
+                    ros_navigation_client_config.clone(),
+                )) as Arc<dyn Navigation>)
+            } else if config.use_navigation_urdf_viz_web_client {
+                let urdf_viz_client = Arc::new(UrdfVizWebClient::default());
+                urdf_viz_client.run_thread();
+                Some(urdf_viz_client as Arc<dyn Navigation>)
+            } else {
+                None
+            };
```
OTL, line 120: I want to refactor these like below

```rust
let navigation = create_navigation(xxx);
```

```rust
fn create_navigation(xxx) {
  #[cfg(feature = "ros")]
  create_navigation_with_ros();
  #[cfg(not(feature = "ros"))]
  create_navigation_without_ros();
}
```

But we have similar structure already, let's do it another change....

OTL (approved):

openrr-apps/src/robot_client.rs
mitsuharu-kojima: I will move these feature(ros) dependent code out of robot client.

PR #79, opened 2021-01-21
[arci] Create DummyNavigation struct

ssr-yuki:
This PR adds DummyNavigation struct, which is based on DummyMoveBase struct.


OTL (approved):
Thanks!

OTL (approved):

PR #84, opened 2021-01-22
Refactor robot client

mitsuharu-kojima:


openrr-client/src/robot_client.rs
```diff
@@ -338,3 +337,37 @@ where
...
+            path.as_ref()
+                .parent()
+                .ok_or_else(|| Error::NoParentDirectory(path.as_ref().to_owned()))?
+                .join(&self.urdf_path),
+        );
+        Ok(())
+    }
+    pub fn urdf_full_path(&self) -> &Option<PathBuf> {
+        &self.urdf_full_path
+    }
```
OTL, line 365: Comment:

I thought that it is better to generate `udf_full_path` if there are no path yet.

```rust
pub fn urdf_full_path(&self) -> Result<&PathBuf> {
    if self.urdf_full_path.is_none() {
       self.resolve_path(&self.urdf_path)?;
    }
    &self.urdf_full_path
}
```

OTL (approved):

PR #85, opened 2021-01-22
Testing arci urdf viz

a-miyadera:
Add tests for async functions


a-miyadera:
Result of testing UrdfVizWebClient::run_thread()

- Quit test function soon after calling run_thread() -> test passed
- Wait 1sec after calling run_thread() before quit test function -> test passed, but thread panicked (see also below)
> thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Status(400, Response[status: 400, status_text: Bad Request])', arci-urdf-viz/src/client.rs:84:48

OTL (approved):

PR #87, opened 2021-01-22
Openrr command

mitsuharu-kojima:


OTL (approved):
👍🏼
Comment:
We need rename the files to the same names with the struct names.

PR #109, opened 2021-01-27
[openrr-command] Keep sending velocity in send_base_velocity command

mitsuharu-kojima:


openrr-command/src/robot_command.rs
```diff
@@ -328,11 +328,13 @@ impl RobotCommandExecutor {
...
                 y,
                 theta,
-                wait_secs,
+                duration_secs,
             } => {
-                client.send_velocity(&BaseVelocity::new(*x, *y, *theta))?;
-                // wait publish
-                sleep(Duration::from_secs_f64(*wait_secs));
+                let start = SystemTime::now();
+                while SystemTime::now() < (start + Duration::from_secs_f64(*duration_secs)) {
```
OTL, line 334: It's better to use `Instant` (and `elapsed()`) in this case. Instant is monotonic clock.

https://doc.rust-lang.org/std/time/struct.Instant.html

`SystemTime` can be go back to the past.

openrr-command/src/robot_command.rs
```diff
@@ -328,11 +328,13 @@ impl RobotCommandExecutor {
...
                 y,
                 theta,
-                wait_secs,
+                duration_secs,
             } => {
-                client.send_velocity(&BaseVelocity::new(*x, *y, *theta))?;
-                // wait publish
-                sleep(Duration::from_secs_f64(*wait_secs));
+                let start = Instant::now();
+                while start.elapsed() < Duration::from_secs_f64(*duration_secs) {
```
OTL, line 334: This is a small improvement, but i want to move `Duration::from` out of the loop. 
`Duration::from_secs_f64(*duration_secs)` is called so many times (for example, 300times for 3secss).
and L336.

OTL (approved):

Issue #35, opened 2021-01-04
Add licenses in all source files


OTL:

taiki-e:
As far as I know, per-file license notifications are not always necessary. Actually, rust-lang/rust removed it from all files: https://github.com/rust-lang/rust/pull/57108

However, it is preferable that all published crates have a license file. This is required when redistributing crates. See https://github.com/mmastrac/rust-ctor/issues/109 for more.

OTL:
I see, thanks!
I close this issue.

PR #111, opened 2021-01-27
Add license file to all crates

taiki-e:
See https://github.com/openrr/openrr/issues/35#issuecomment-768085589


arci-ros/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
OTL, line 1: `arci-ros` includes the file which is from BSD lisences in arci-ros/ros_msgs dir.
It contains LICENSE.BSD, which I created.

Do you think that is it safe to say that this is Apache-2?
Shouldn't we move LICENSE.BSD to `arci-ros/`?

OTL (approved):

arci-ros/LICENSE
taiki-e: Hmm, it seems LICENSE.BSD is BSD 2-clause, so I think it's okay to distribute it as part of the Apache-2 licensed software. (Disclaimer: I am not a lawyer) 
However, I think we need to add a file like rust-lang/rust's [COPYRIGHT](https://github.com/rust-lang/rust/blob/78e22069d018e83915201c8a218a0a94227f6420/COPYRIGHT) to the arci-ros crate regarding the inclusion of source files with different licenses.

refs: 
- https://www.apache.org/legal/resolved.html (BSD 2-clause is in category A)
- https://infra.apache.org/licensing-howto.html (see "Bundling permissively-licensed dependencies" section)

OTL: I see. Anyway this PR improves current situation, let's merge this.

PR #113, opened 2021-01-28
[arci-urdf-viz] send_joint_positions in one thread

mitsuharu-kojima:
- to avoid multiple spawning of send_joint_positions thread.
- to follow latest target.

### DIFF
--- a/arci-urdf-viz/src/client.rs
+++ b/arci-urdf-viz/src/client.rs
@@ -4,12 +4,18 @@ use arci::{
     Navigation, SetCompleteCondition, TotalJointDiffCondition,
 };
 use async_trait::async_trait;
+use log::debug;
 use nalgebra as na;
 use openrr_sleep::ScopedSleep;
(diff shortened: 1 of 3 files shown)


arci-urdf-viz/src/client.rs
```diff
@@ -84,6 +103,77 @@ impl UrdfVizWebClient {
             send_robot_origin(&base_url, pose).unwrap();
         });
     }
+    pub fn run_send_joint_positions_thread(&mut self) {
+        if self.send_joint_positions_thread.is_some() {
+            panic!("send_joint_positions_thread is running.");
+        }
```
OTL, line 109: If we can make this private, and automatically runs in the cunstructor, we can remove this check and avoid `panic!`. Is it possible?

OTL: Or use scoped thread (crossbeam)?

OTL (approved):

PR #114, opened 2021-01-28
Correct name of ignored directory

ssr-yuki:
In `.gitignore`, `/target/` notation just ignores `openrr/target`, however it does not other `target` directories in other hierarchies, such as `openrr/arci/target`. This PR fixes the problem.


OTL:
I don't think that `arci/target` is generated, because we uses workspace.
All binaries should be located in `target/` even if you `cargo build` in `arci`.

I want to know why you are in such situation.

OTL:
It seems that `target` or `target/` is more popular.

https://github.com/rust-lang/cargo/blob/master/.gitignore
https://github.com/rust-lang/futures-rs/blob/master/.gitignore
https://github.com/crossbeam-rs/crossbeam/blob/master/.gitignore

OTL (approved):

ssr-yuki:
Thanks!

PR #117, opened 2021-01-28
[openrr-apps, openrr-teleop] Support JointsPose in robot_teleop

mitsuharu-kojima:


openrr-teleop/src/joints_pose_sender.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+impl<S> JointsPoseSender<S>
+where
+    S: Speaker,
+{
+    pub fn new(
+        mode: String,
+        joints_poses: Vec<JointsPose>,
+        all_joint_trajectory_clients: &HashMap<String, Arc<dyn JointTrajectoryClient>>,
```
OTL, line 33: I don't want to introduce the structure of `RobotClient` here.
This should be independent from it as a design.

I think it is not natural to create `JointsPoseSender` with all_joint_trajectory_clients.

OTL (approved):
🙏🏻🙏🏻🙏🏻

PR #118, opened 2021-01-28
[openrr-client] Add LocalMove

ssr-yuki:
This PR adds a struct to move robots to a goal directly.

The placement of the sample configuration file named `local_move_sample.toml`, which is used for tests, should be discussed.

### DIFF
--- a/openrr-client/src/clients/local_move.rs
+++ b/openrr-client/src/clients/local_move.rs
@@ -51,12 +51,12 @@ impl LocalMove {
             && pose_error.rotation.angle().abs() < self.config.reach_angle_threshold
     }
 
-    pub fn send_zero_vel(&self) -> Result<(), arci::Error> {
+    pub fn send_zero_velocity(&self) -> Result<(), arci::Error> {
         self.vel_client
             .send_velocity(&BaseVelocity::new(0.0, 0.0, 0.0))
     }
 
-    pub fn send_control_vel_from_pose_error(
+    pub fn send_control_velocity_from_pose_error(
         &self,
         pose_error: Isometry2<f64>,
     ) -> Result<(), arci::Error> {
@@ -127,7 +127,7 @@ mod tests {
 
         // Set control velocity
         local_move
-            .send_control_vel_from_pose_error(Isometry2::new(Vector2::new(1.0, -1.0), 1.0))
+            .send_control_velocity_from_pose_error(Isometry2::new(Vector2::new(1.0, -1.0), 1.0))
             .unwrap();
         let vel0 = local_move.vel_client.current_velocity().unwrap();
         assert_eq!(vel0.x, 1.0); // linear_gain = 1.0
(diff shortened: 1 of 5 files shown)


openrr-client/src/clients/local_move.rs
```diff
@@ -0,0 +1,152 @@
...
+            config,
+        }
+    }
+
+    pub fn is_reached(&self, pose_error: Isometry2<f64>) -> bool {
+        pose_error.translation.vector.norm() < self.config.reach_distance_threshold
+            && pose_error.rotation.angle().abs() < self.config.reach_angle_threshold
+    }
+
+    pub fn send_zero_vel(&self) -> Result<(), arci::Error> {
```
OTL, line 54: I feel `send_zero_velocity` is better.

openrr-client/src/error.rs
```diff
@@ -5,22 +5,26 @@ use urdf_rs::UrdfError;
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum Error {
-    #[error("openrr-cleint: No JointTrajectoryClient={} is found.", .0)]
-    NoJointTrajectoryClient(String),
+    #[error("openrr-cleint: arci: {:?}", .0)]
```
OTL, line 8: Do you know why this soring happed?

openrr-client/src/clients/local_move.rs
```diff
@@ -0,0 +1,152 @@
...
+                .map_err(|e| Error::NoFile(path.as_ref().to_owned(), e))?,
+        )
+        .map_err(|e| Error::TomlParseFailure(path.as_ref().to_owned(), e))?;
+        Ok(config)
+    }
+}
+
+pub struct LocalMove {
+    pub nav_client: Box<dyn Navigation>,
+    pub vel_client: Box<dyn MoveBase>,
```
OTL, line 32: Could you try these generics instead of Box<dyn XXX>?

In many cases Box<dyn XXX> implements XXX automatically by #[auto_Impl].

OTL (commented):
What is the plan to implement the control loop?
I want to use this for dummy navigation in urdf-viz.

openrr-client/src/error.rs
ssr-yuki: I did it manually, since I cannot find any orders to define `Errors`. 

If you have a problem, I'll revert it.

openrr-client/src/clients/local_move.rs
ssr-yuki: Ok, I try it.

openrr-client/src/error.rs
OTL: If it is your intention (not a mistake, merge conflict nor something), it's OK.

ssr-yuki:
> What is the plan to implement the control loop?
> I want to use this for dummy navigation in urdf-viz.

I want to discuss it with you. I consider the following two candidates:
1. implement a function for the control loop
2. NOT Implement such a function but implement a binary whose `main` calls LocalMove functions in while loops.

I think both of them are acceptable; in my opinion, implementing the control loop as a struct method is not good since users cannot add break rules in their own cases. I worry about that, e.g. in case actuators do not work, such a function falls into infinite loop with publishing non-zero commands.

Please give me comments on this.

openrr-client/src/clients/local_move.rs
ssr-yuki: Ok, and I did it.

ssr-yuki:
I think implementing the control loop requires time measurement (e.g. using `arci_ros::rate()`), and `LocalMove` struct should not include it.

OTL (approved):

PR #125, opened 2021-02-01
[arci-speak-cmd] Add arci-speak-cmd

### LINKED ISSUE #119: Implement local command arci::Speak
arci::Speak can be implemented as a local command.

We can use bellow commands depends on the OS.

* MacOS: `say`
* Linux: `espeak` (not default but easy to use)
* Windows: I don't know!

taiki-e:
Closes #119

### APIs

```rust
/// A [`Speaker`] implementation using a local command.
///
/// Currently, this uses the following command:
///
/// - On macOS, use `say` command.
/// - On Windows, call [SAPI] via PowerShell.
/// - On others, use `espeak` command.
///
/// **Disclaimer**: These command might change over time.
///
/// [SAPI]: https://en.wikipedia.org/wiki/Microsoft_Speech_API
#[derive(Debug, Default)]
pub struct LocalCommand { /* ... */ }

impl LocalCommand {
    /// Creates a new `LocalCommand`.
    pub fn new() -> Self { /* ... */ }

    /// Similar to `Speaker::speak`, but returns an error when the command failed.
    pub fn try_speak(&self, message: &str) -> io::Result<()> { /* ... */ }
}

impl Speaker for LocalCommand { /* ... */ }
```

### Example

```rust
use arci::Speaker;
use arci_speak_cmd::LocalCommand;
use structopt::StructOpt;

#[derive(Debug, StructOpt)]
struct Args {
    message: String,
}

fn main() {
    env_logger::init();
    let args = Args::from_args();

    let speaker = LocalCommand::default();
    speaker.speak(&args.message)
}
```

```
$ cargo run -p arci-speak-cmd --example local_command -- "hello"
```

### DIFF
--- a/arci-speak-cmd/src/lib.rs
+++ b/arci-speak-cmd/src/lib.rs
@@ -1,0 +1,76 @@
+use arci::Speaker;
+use log::error;
+use std::{io, process::Command};
+
+/// A [`Speaker`] implementation using a local command.
+///
+/// Currently, this uses the following command:
+///
+/// - On macOS, use `say` command.
+/// - On Windows, call [SAPI] via PowerShell.
+/// - On others, use `espeak` command.
+///
+/// **Disclaimer**: These commands might change over time.
+///
+/// [SAPI]: https://en.wikipedia.org/wiki/Microsoft_Speech_API
+#[derive(Debug, Default)]
+#[non_exhaustive]
+pub struct LocalCommand {}
+
+impl LocalCommand {
+    /// Creates a new `LocalCommand`.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Similar to `Speaker::speak`, but returns an error when the command failed.
+    pub fn try_speak(&self, message: &str) -> io::Result<()> {
+        run_local_command(message)
+    }
+}
+
+impl Speaker for LocalCommand {
+    fn speak(&self, message: &str) {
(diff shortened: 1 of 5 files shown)


arci-speak-cmd/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+            // TODO: Speaker trait seems to assume that speak method will always succeed.
+            error!("{}", e);
+        }
+    }
+}
+
+#[cfg(not(windows))]
+fn run_local_command(message: &str) -> io::Result<()> {
+    #[cfg(not(target_os = "macos"))]
+    let cmd_name = "espeak";
```
OTL, line 44: Isn't is better to use `const CMD_NAME: &str = "..."`?

arci-speak-cmd/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    let mut cmd = Command::new(cmd_name);
+    let status = cmd.arg(message).status()?;
+
+    if status.success() {
+        Ok(())
+    } else {
+        Err(io::Error::new(
+            io::ErrorKind::Other,
+            format!("failed to run `{}`", cmd_name),
```
OTL, line 56: I want to include `message` in this message.
L73, too.

OTL (approved):

OTL (approved):
👍🏼👍🏼👍🏼

PR #150, opened 2021-02-09
[workflows] Add openrr_apps ros1 job

mitsuharu-kojima:


mitsuharu-kojima:
@OTL  ros1 check will never complete with current setting,  If you approved this change, I will update the setting.

OTL:
Oh, let me check!

OTL (approved):

PR #180, opened 2021-02-17
[openrr-client, openrr-apps] Make urdf path and clients configs optional

mitsuharu-kojima:
- to use robot client / config only with navigation


OTL (approved):
👍🏼

PR #184, opened 2021-02-19
[CI] Fix coverage

OTL:
* grcov v0.7.1 causes problems for now, so use v0.6.1
* Use source-based coverage

https://github.com/mozilla/grcov/issues/555
https://github.com/mozilla/grcov/issues/575


OTL:
I tried a lot, but i think grcov for source-based coverage has so much bugs yet. I gave up this!

PR #187, opened 2021-02-19
[arci-speak-cmd] v0.1.0 -> v0.0.3

OTL:
It should have been 0.0.1, but v0.1.0 is released.
When we release v0.1.0, we have to skip v0.1.0
for arci-speak-cmd.


OTL:
refs #191

PR #201, opened 2021-02-25
[openrr-teleop, openrr-apps] Make initial teleop mode configurable

mitsuharu-kojima:


openrr-apps/src/bin/robot_teleop.rs
```diff
@@ -41,7 +41,22 @@ async fn main() -> Result<(), Error> {
...
-    let switcher = Arc::new(ControlNodeSwitcher::new(nodes, client.clone()));
+    let initial_node_index = if teleop_config.initial_mode.is_empty() {
+        0
+    } else if let Some(initial_node_index) = nodes
+        .iter()
+        .position(|node| node.mode() == teleop_config.initial_mode)
+    {
+        initial_node_index
+    } else {
+        0
```
OTL, line 52: I think it's better to add warning to show that teleop_config.initial_mode is not found.
Or It might be better to die.

OTL (approved):
🙏🏻🙏🏻🙏🏻

PR #207, opened 2021-02-25
[workflows] Separate the dependency installation on linux into a script

taiki-e:
If we forget to update the dependency installation command for the Linux environment, #206 will break easily. And it's hard to notice until the release process actually started.

This patch separates the dependency installation commands into scripts for easier management. Also, this makes workflows that aren't run regularly like #206 hard to break.

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -18,9 +18,7 @@ jobs:
         with:
           toolchain: nightly
           override: true
-      - run: sudo apt-get update
-      # only for arci-gamepad-gilrs
-      - run: sudo apt-get install -y libudev-dev libasound2-dev
+      - run: ci/ubuntu-install-dependencies.sh
       # only for arci-ros test (roscore)
       - run: sudo apt-get install -y python-roslaunch python-rostopic
       - uses: actions-rs/cargo@v1
--- a/ci/ubuntu-install-dependencies.sh
+++ b/ci/ubuntu-install-dependencies.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+sudo apt-get update
+
+# libudev-dev is for arci-gamepad-gilrs
+# libasound2-dev is for arci-speak-audio
+sudo apt-get install -y \
+    libudev-dev libasound2-dev
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,12 +58,11 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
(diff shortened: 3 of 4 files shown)


ci/ubuntu-install-dependencies.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+sudo apt-get update
+
+# only for arci-gamepad-gilrs
+sudo apt-get install -y libudev-dev libasound2-dev
```
OTL, line 9: libasound2-dev is for arci-speak-audio (The original comment was already wrong).
Could you update the comment?

.github/workflows/coverage.yaml
```diff
@@ -18,9 +18,7 @@ jobs:
         with:
           toolchain: nightly
           override: true
-      - run: sudo apt-get update
-      # only for arci-gamepad-gilrs
-      - run: sudo apt-get install -y libudev-dev libasound2-dev
+      - run: ci/ubuntu-install-dependencies.sh
       # only for arci-ros test (roscore)
       - run: sudo apt-get install -y python-roslaunch python-rostopic
```
taiki-e, line 23: The installations of python-roslaunch and python-rostopic don't work well in `ros:melodic-ros-base` images, so they're leaved for now.

https://github.com/openrr/openrr/runs/1977489062?check_suite_focus=true

```text
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 python-roslaunch : Depends: python-roslib but it is not going to be installed
 python-rostopic : Depends: python-rospy but it is not going to be installed
                   Depends: python-rosbag but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

OTL (approved):

ci/ubuntu-install-dependencies.sh
taiki-e: Updated the comment.

.github/workflows/coverage.yaml
OTL: I think if you are using ros:melodic-ros-base, they have been already installed. So you can skip it.
(Add some option to skip?)

OTL: ros:melodoc-ros-base means that `(base of )ROS is (fully) installed`, and I want to test on `NOT ROS installed` here.(but needs a small part of ROS for testing)
I mean ros:melodoc-ros-base should not be used basicly, and keep this as you did.

OTL (approved):

Issue #101, opened 2021-01-26
openrr-gui: Improvement of joint_position_sender


taiki-e:
Bugfixes:

- TODO

Resolve TODOs:

- TODO

OTL:
Request:

Add text box(?) to set the duration. (Currently I ordered to use 100msec)
The unit of time should be f64 sec.

taiki-e:
We also need to rethink when to apply the joint position text box updates.
Perhaps it is preferable to request an explicit enter key (using [iced::TextInput::on_submit](https://docs.rs/iced_native/0.3.0/iced_native/widget/text_input/struct.TextInput.html#method.on_submit)). 
Currently, it is updated every time you type a new character. (For example, if you try to type 1.35, joint_position_sender will send 1, 1.3, and 1.35.)

PR #210, opened 2021-02-26
[openrr-gui] Add text input to set duration and improve error handling about text input

taiki-e:
cc https://github.com/openrr/openrr/issues/101


taiki-e:
https://user-images.githubusercontent.com/43724913/109290713-505c6b80-786b-11eb-8df4-57798da0b059.mov

OTL (approved):

PR #211, opened 2021-03-02
[arci] Change Speaker trait interface

### LINKED ISSUE #174: Change Speak trait interface
Current

```rust
pub trait Speaker: Send + Sync {
    fn speak(&self, message: &str);
}
```

Proposed

```rust
pub trait Speaker: Send + Sync {
    async fn speak(&self, message: &str) -> Result<()>;
}
```

It should be `async` because it takes sometime to finish speaking.
And it might be failed.

taiki-e:
- Change to return `Result<(), arci::Error>`
- Change `Speaker::speak` to async method

Closes #174

### DIFF
--- a/openrr-teleop/src/control_node.rs
+++ b/openrr-teleop/src/control_node.rs
@@ -5,7 +5,7 @@ use auto_impl::auto_impl;
 #[async_trait]
 #[auto_impl(Box)]
 pub trait ControlNode: Send + Sync {
-    fn set_event(&mut self, event: GamepadEvent);
+    async fn set_event(&mut self, event: GamepadEvent);
     async fn proc(&self);
     fn mode(&self) -> &str;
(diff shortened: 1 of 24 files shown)


openrr-teleop/src/control_node.rs
```diff
@@ -5,7 +5,7 @@ use auto_impl::auto_impl;
 #[async_trait]
 #[auto_impl(Box)]
 pub trait ControlNode: Send + Sync {
-    fn set_event(&mut self, event: GamepadEvent);
+    async fn set_event(&mut self, event: GamepadEvent);
```
taiki-e, line 8: Notice that `ControlNode::set_event` also changed to async because some implementers are calling `Speaker::speak`.

taiki-e:
blocked on #228, so closing this for now.

PR #214, opened 2021-03-03
Add vendor and .cargo to gitignore

taiki-e:
You can develop offline by vendoring all dependencies locally
More information here https://doc.rust-lang.org/cargo/commands/cargo-vendor.html


taiki-e:
Hmm, it seems I can build offline without vendor...

PR #230, opened 2021-03-25
[arci-urdf-viz, openrr-gui] Accept `duration = 0.0`

taiki-e:


arci-urdf-viz/src/client.rs
```diff
@@ -140,6 +140,20 @@ impl UrdfVizWebClient {
                         .unwrap()
                         .positions;
                     let duration_sec = target.duration.as_secs_f64();
+
+                    if duration_sec == 0.0 {
```
OTL, line 144: I feel bad when compare float using '=='.

I think it is better to use 'is_zero()' to duration instance instead of f64.

```suggestion
                    if target.duration.is_zero() {
```

I know that it's not problem actually.

openrr-gui/src/joint_position_sender.rs
```diff
@@ -412,7 +412,7 @@ where
                         // We don't round the input for now. If we do that, we also need to update
                         // the text input that we show to the user.
 
-                        if !duration.is_normal() {
+                        if !duration.is_normal() && duration != 0.0 {
```
OTL, line 415: same as above, i don't want to compare using '!='.
And it should be positive number.

How about

```suggestion
                        if !duration.is_normal() || duration.is_sign_negative() {
```

or something?

taiki-e, line 415:
```suggestion
                        if !duration.is_normal() && !duration.is_sign_negative() {
```

OTL: You are right!

taiki-e:
🤔 

```text
error[E0658]: use of unstable library feature 'duration_zero'
   --> arci-urdf-viz/src/client.rs:136:40
    |
136 |                     if target.duration.is_zero() {
    |                                        ^^^^^^^
    |
    = note: see issue #73544 <https://github.com/rust-lang/rust/issues/73544> for more information
```

arci-urdf-viz/src/client.rs
```diff
@@ -133,6 +133,19 @@ impl UrdfVizWebClient {
                 let send_joint_positions_target =
                     { send_joint_positions_target_arc_mutex.lock().unwrap().take() };
                 if let SendJointPositionsTargetState::Some(target) = send_joint_positions_target {
+                    if target.duration.as_nanos() == 0 {
```
taiki-e, line 136: workaround for "use of unstable library feature 'duration_zero'"

OTL (approved):

arci-urdf-viz/src/client.rs
OTL: Sorry!

OTL:
merged

PR #243, opened 2021-04-01
Go back to sync

### LINKED ISSUE #228: Go back to sync
It seems that async is not used effectively,  and it has weak reason to use async for now.
Let's go back to old implementation.

Return boxed Wait instead of Future.

```rust
use crate::error::Error;

pub type Wait<'a> = Box<dyn WaitTrait + 'a>;

pub trait WaitTrait {
    fn wait(&self) -> Result<(), Error>;
}
```

Current

```rust
a.send_joint_positions()?.await?;
```

Go back to

```rust
a.send_joint_positions()?.wait()?;
```

taiki-e:
**Work in progress**

Closes #228


taiki-e:
closing in favor of #248

PR #244, opened 2021-04-01
Use assert_approx_eq for float comparison

taiki-e:
```console
$ cargo clippy --all-targets
error: strict comparison of `f32` or `f64`
   --> openrr-client/src/clients/local_move.rs:139:9
    |
139 |         assert_eq!(vel0.x, 1.0); // linear_gain = 1.0
    |         ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::float_cmp)]` on by default
    = note: `f32::EPSILON` and `f64::EPSILON` are available for the `error_margin`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#float_cmp
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #246, opened 2021-04-02
[workflows] Apply clippy for tests and examples

taiki-e:
Addresses https://github.com/openrr/openrr/pull/244#pullrequestreview-626733552

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -114,4 +114,4 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
+          args: --all-targets -- -D warnings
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -114,4 +114,4 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
+          args: --all-targets -- -D warnings
```
taiki-e, line 117: `--all-targets`'s "target" means lib, bin, example, test, bench, etc., not `--target`'s "target (target triple)".

OTL (approved):
👍🏼👍🏼👍🏼

PR #261, opened 2021-04-12
[arci, openrr-client] Add move_joint_until_stop

mitsuharu-kojima:


arci/src/lib.rs
```diff
@@ -1,11 +1,13 @@
 mod clients;
 mod error;
 mod traits;
+mod utils;
 mod waits;
 
 pub use clients::*;
 pub use error::*;
 pub use traits::*;
+pub use utils::*;
```
OTL, line 10: I prefer not to export the functions in utils into the root and want to export the `utils` module.

```suggestion
pub mod utils;
mod waits;

pub use clients::*;
pub use error::*;
pub use traits::*;
```

For example, `urdf-rs`
https://github.com/openrr/urdf-rs/blob/main/src/lib.rs

What do you think?

OTL (approved):

PR #272, opened 2021-04-13
[arci-urdf-viz] Fix flaky test

taiki-e:
TotalJointDiffCondition allows error of 0.02, so WaitFuture may complete before the joint reaches the target. So, allow more error than TotalJointDiffCondition allows, when comparison.


OTL (approved):
🙏🏻🙏🏻🙏🏻

PR #274, opened 2021-04-14
add test for JointTrajectoryClientsContainer

chatblanc-ciel:
- ``fn current position``
  - ok pattern
  - err pattern

- ``fn send position``
  - ok pattern
  - err pattern

Remaking pull request.

### DIFF
--- a/arci/src/clients/joint_trajectory_clients_container.rs
+++ b/arci/src/clients/joint_trajectory_clients_container.rs
@@ -1,7 +1,10 @@
-use crate::error::Error;
-use crate::traits::{JointTrajectoryClient, TrajectoryPoint};
-use crate::waits::WaitFuture;
 use futures::stream::FuturesOrdered;
+
+use crate::{
+    error::Error,
+    traits::{JointTrajectoryClient, TrajectoryPoint},
+    waits::WaitFuture,
+};
 
 pub struct JointTrajectoryClientsContainer<T: JointTrajectoryClient> {
     joint_names: Vec<String>,
@@ -31,6 +34,7 @@ where
     fn joint_names(&self) -> &[String] {
         &self.joint_names
     }
+
     fn current_joint_positions(&self) -> Result<Vec<f64>, Error> {
         let mut ret = vec![];
         for c in &self.clients {
@@ -39,6 +43,7 @@ where
         }
         Ok(ret)
     }
+
     fn send_joint_positions(
         &self,
         positions: Vec<f64>,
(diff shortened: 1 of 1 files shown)


arci/src/clients/joint_trajectory_clients_container.rs
```diff
@@ -207,4 +207,211 @@ mod tests {
...
+        let container = JointTrajectoryClientsContainer::new(clients);
+        let current = container.current_joint_positions();
+        let expect = vec![1.0_f64, 3.0_f64, 2.2_f64, 4.0_f64];
+
+        let positions = current.unwrap();
+
+        positions
+            .iter()
+            .zip(expect)
+            .for_each(|(pos, correct)| assert_approx_eq!(*pos, correct));
```
OTL, line 318: I think it's better to check current is error, I mean

```rust
assert!(current.is_err());
```

and remove L315-318, L265.
What do you think?

arci/src/clients/joint_trajectory_clients_container.rs
```diff
@@ -207,4 +207,211 @@ mod tests {
...
+            Dummy {
+                name: vec![String::from("part2"), String::from("low")],
+                pos: vec![2.2_f64, 4.0_f64],
+            },
+        ];
+
+        let container = JointTrajectoryClientsContainer::new(clients);
+        let pos = vec![3.4_f64, 5.8_f64, 0.1_f64, 2.5_f64];
+        let duration = std::time::Duration::from_secs(5);
+        let _ = container.send_joint_positions(pos, duration).unwrap();
```
OTL, line 365: I want to use `await` to check wait logic is working.

```suggestion
        container.send_joint_positions(pos, duration).unwrap().await;
```

To do that, we need `#[tokio::test]` and `async` to the function name.
You can see the examples [here](https://github.com/openrr/openrr/blob/main/arci/tests/test_waits.rs)

arci/src/clients/joint_trajectory_clients_container.rs
chatblanc-ciel: I agree with your purpose.
I'll change it.

In other view, ``unwrap`` is often using case.
So, is ``unwrap`` test necessary?

arci/src/clients/joint_trajectory_clients_container.rs
chatblanc-ciel: Is ` assert!(container.send_joint_positions(pos, duration).is_err());` bat test?

I want to know the difference between `unwrap test` and ` unwrap + await`.

OTL (approved):

OTL (approved):

PR #276, opened 2021-04-16
[arci-ros] Handle wait-related error in RosNavClient::send_goal_pose

### LINKED ISSUE #172: RosNavClient::send_pose() return Ok even if timeout!!
It's not clear what should send_pose() returns, but i think it's a bug.

- [ ] Make it clear what should send_pose() returns
- [ ] fix it

In the #144, there are discuss about the naming of `send_pose()`

```
Navigation::send_pose() -> Navigation::send_goal_pose()
```

`send_pose()` is not `send_goal_pose()`.
It's actually `move_to`.

taiki-e:
Fixes #172


OTL (approved):
Thanks!

PR #279, opened 2021-04-20
Add rustfmt.toml to unify format

mitsuharu-kojima:


OTL:
The purpose of this change is to get stable results with stable and nightly rustfmt, right?
stable rustfmt is changed?

OTL (approved):

PR #282, opened 2021-04-20
[arci] Add JointPositionLimiter

taiki-e:
- [x] arci: add JointPositionLimiter
- [x] arci-urdf-viz: use JointPositionLimiter
- [x] arci-ros: use JointPositionLimiter


arci/src/clients/joint_position_limiter.rs
```diff
@@ -0,0 +1,362 @@
...
+        }
+        self.client.send_joint_trajectory(trajectory)
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum JointPositionLimiterStrategy {
+    /// If the position is out of the limit, handle it as the same value as the limit.
+    Round,
```
OTL, line 143: I don't know about English but we use `clamp` for this.

https://docs.rs/k/0.24.0/k/struct.Chain.html#method.set_joint_positions_clamped

arci/src/clients/joint_position_limiter.rs
```diff
@@ -0,0 +1,362 @@
...
+impl<C> JointTrajectoryClient for JointPositionLimiter<C>
+where
+    C: JointTrajectoryClient,
+{
+    fn joint_names(&self) -> &[String] {
+        self.client.joint_names()
+    }
+
+    fn current_joint_positions(&self) -> Result<Vec<f64>, Error> {
+        self.client.current_joint_positions()
```
OTL, line 116: This may be optional, but it might be a good idea to use `check_joint_position()` here.
In the real world, the robot returns the values out of the limits, and the system can't handle the values which are out of the limit. If this handles this, it reduces the cost to check the current_joint_positions in the user logics.

arci/src/clients/joint_position_limiter.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub fn from_urdf_with_strategy(
+        client: C,
+        joints: &[urdf_rs::Joint],
+        strategy: JointPositionLimiterStrategy,
+    ) -> Result<Self, Error> {
+        let mut limits = Vec::new();
+        for joint_name in client.joint_names() {
+            if let Some(i) = joints.iter().position(|j| j.name == *joint_name) {
+                let joint = &joints[i];
+                let limit = if JointType::Continuous == joint.joint_type {
```
OTL, line 55: If the joint is continuous, it should not have limits.

See
https://wiki.ros.org/urdf/XML/joint

```
<limit> (required only for revolute and prismatic joint)
```

arci/src/clients/joint_position_limiter.rs
taiki-e: done in d01a272

arci/src/clients/joint_position_limiter.rs
taiki-e: done in 1a5d5e1

arci/src/clients/joint_position_limiter.rs
taiki-e: done in 807d711

OTL (approved):

mitsuharu-kojima (approved):

PR #285, opened 2021-04-22
Group exports from modules

taiki-e:
https://github.com/openrr/openrr/blob/1dc6dfe09c82e91b21a9c3ce1e6aae887276e924/arci/src/lib.rs#L9-L16

### DIFF
--- a/arci/src/lib.rs
+++ b/arci/src/lib.rs
@@ -8,9 +8,6 @@ mod waits;
 
 // re-export
 pub use async_trait::async_trait;
-pub use clients::*;
-pub use error::*;
-// re-export
 pub use nalgebra::{Isometry2, Isometry3};
-pub use traits::*;
-pub use waits::*;
+
+pub use crate::{clients::*, error::*, traits::*, waits::*};
--- a/arci-ros/src/lib.rs
+++ b/arci-ros/src/lib.rs
@@ -13,15 +13,11 @@ mod ros_speak_client;
 pub mod ros_transform_resolver;
 pub mod rosrust_utils;
 
-pub use cmd_vel_move_base::*;
-pub use error::Error;
-pub use joy_gamepad::*;
-pub use ros_control_action_client::*;
-pub use ros_control_client::*;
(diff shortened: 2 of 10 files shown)


arci/src/lib.rs
```diff
@@ -10,7 +10,6 @@ mod waits;
 pub use async_trait::async_trait;
 pub use clients::*;
 pub use error::*;
-// re-export
 pub use nalgebra::{Isometry2, Isometry3};
 pub use traits::*;
 pub use waits::*;
```
OTL, line 15: re-exporting only `async_trait` and `nalgebra`.
Others are inside of this crate.

How about sorting?

```rust
pub use clients::*;
pub use error::*;
pub use traits::*;
pub use waits::*;

// re-export
pub use async_trait::async_trait;
pub use nalgebra::{Isometry2, Isometry3};
```

mitsuharu-kojima: FYI:
> How about sorting?
The order is forced by the rustfmt setting.

OTL: Oh, I understand that it's not possible to sort!
(I wanted to add comments about these lines are exporting **external crates**)

taiki-e: > I wanted to add comments about these lines are exporting external crates

Oh, I didn't notice it was intentional.

> Currently the order is forced by the rustfmt setting.

FWIW, the workaround for this rustfmt issue is to add the `crate::` or `self::` prefix. It will be formatted as follows:

```rust
// re-export
pub use async_trait::async_trait;
pub use nalgebra::{Isometry2, Isometry3};

pub use crate::{clients::*, error::*, traits::*, waits::*};
```

refs: https://github.com/rust-lang/rustfmt/issues/4709

OTL: I think we can use your proposal (use crate::).

taiki-e: Ok, updated to group exports from modules.

OTL (approved):

PR #289, opened 2021-04-26
[workflows] Fix installation of rustfmt

taiki-e:
https://github.com/openrr/openrr/runs/2434145731?check_suite_focus=true

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -93,7 +93,7 @@ jobs:
           profile: minimal
           toolchain: nightly
           override: true
-      - run: rustup component add rustfmt
+          components: rustfmt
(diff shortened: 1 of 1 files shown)


taiki-e:
refs: https://github.com/actions-rs/toolchain#components

> As an extra perk, rustup >= 1.20.0 is able to find the most recent nightly toolchain with the requested components available; next example is utilizing this feature to install the minimal set of nightly toolchain components with the rustfmt and clippy extras:

> In case if nightly toolchain is requested and one of the components is missing in latest nightly release, this Action will attempt the downgrade till it find the most recent nightly with all components needed.

OTL (approved):
👍🏼👍🏼👍🏼

PR #293, opened 2021-04-28
[arci-gamepad-gilrs] Fix type infer error

taiki-e:
https://github.com/openrr/openrr/runs/2456586013?check_suite_focus=true

    error[E0283]: type annotations needed
      --> arci-gamepad-gilrs/src/lib.rs:231:27
        |
    231 |                     if id == connected_id.into() {
        |                           ^^ ------------------- this method call resolves to `T`
        |                           |
        |                           cannot infer type
        |
        = note: cannot satisfy `usize: PartialEq<_>`


taiki-e:
Oh, this seems to happen only in #292 for some reason. I will do this as part of it.

PR #296, opened 2021-05-06
Make dependency on assimp and mio optional

taiki-e:
Both or either is not available on some platforms, which can be a problem when working on support for platforms like WASM.

This PR makes both dependencies optional features.
This also adds a CI task that checks all feature combinations work properly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-
+resolver = "2"
 members = [
     "arci",
     "arci-gamepad-gilrs",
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-
+resolver = "2"
```
taiki-e, line 2: `resolver = "2"` (that fixes various cargo features/workspace related bugs) is needed to check all feature combinations works properly [without editing `Cargo.toml`](https://github.com/taiki-e/cargo-hack/issues/15).

taiki-e:
I'm not sure if the crates that added the assimp feature in this PR really actually need assimp. (However. if it turns out that it doesn't need it, I don't think that's a big problem since we can remove the feature later.)

OTL (approved):

PR #304, opened 2021-05-12
[openrr-plugin] Add openrr-plugin

### LINKED ISSUE #75: arci plugin system
We want make it a plugin!

Is it possible to do that for async trait?

taiki-e:
Closes #75

## Examples

- [plugin definition](https://github.com/openrr/openrr/blob/taiki-e/plugin/openrr-plugin/examples/plugin/src/lib.rs)
- [plugin loader](https://github.com/openrr/openrr/blob/taiki-e/plugin/openrr-plugin/examples/load_plugin.rs)

### API Design

- [x] Decide on a unit of the plugin.
- [x] Decide what arguments to use when initializing the plugin.
- [x] Decide naming convention of proxy types (R* is confusing.)

### Implementation

- [x] arci::JointTrajectoryClient
- [x] arci::Speaker
- [x] arci::MoveBase
- [x] arci::Navigation
- [x] arci::Localization
- [x] arci::TransformResolver
- [x] arci::Gamepad

### Other

- [x] docs
- [x] tests
- [x] benches

## Future plan

- support openrr-plugin in openrr-apps #312 
- tests more

### DIFF
--- a/openrr-plugin/examples/plugin/Cargo.toml
+++ b/openrr-plugin/examples/plugin/Cargo.toml
@@ -9,6 +9,6 @@ publish = false
 crate-type = ["cdylib"]
 
 [dependencies]
-arci = "0.0.5"
+arci-speak-cmd = "0.0.5"
 openrr-plugin = "0.0.5"
-abi_stable = "0.9.3"
+abi_stable = "0.9"
--- a/openrr-plugin/examples/plugin/Cargo.toml
+++ b/openrr-plugin/examples/plugin/Cargo.toml
@@ -9,6 +9,6 @@ publish = false
 crate-type = ["cdylib"]
 
 [dependencies]
-arci-speak-cmd = "0.0.5"
-openrr-plugin = "0.0.5"
 abi_stable = "0.9"
+openrr-client = "0.0.5"
+openrr-plugin = "0.0.5"
--- a/openrr-plugin/examples/plugin/Cargo.toml
+++ b/openrr-plugin/examples/plugin/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "openrr-plugin-example"
+version = "0.1.0"
+authors = ["Taiki Endo <te316e89@gmail.com>"]
+edition = "2018"
+publish = false
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+arci = "0.0.5"
+openrr-plugin = "0.0.5"
+abi_stable = "0.9.3"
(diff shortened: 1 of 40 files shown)


openrr-plugin/tests/proxy.rs
```diff
@@ -0,0 +1,110 @@
...
+};
+use assert_approx_eq::assert_approx_eq;
+use nalgebra::Vector2;
+use openrr_plugin::{RLocalization, RMoveBase, RNavigation, RSpeaker};
+
+// TODO: move to arci?
+#[derive(Debug, Default)]
+struct DummySpeaker {
+    message: Mutex<String>,
+}
```
taiki-e, line 18: Opened #311

openrr-plugin/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+    ///
+    /// NOTE: This is *not* a unique identifier.
+    fn name(&self) -> String;
+
+    /// Creates a new instance of [`arci::JointTrajectoryClient`] with the specified arguments.
+    fn new_joint_trajectory_client(
+        &self,
+        args: String,
+    ) -> Option<Box<dyn StaticJointTrajectoryClient>> {
+        drop(args);
```
OTL, line 109: Just a question as a rust beginner.
Why we need drop?

openrr-plugin/examples/plugin/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "openrr-plugin-example"
+version = "0.1.0"
+authors = ["Taiki Endo <te316e89@gmail.com>"]
```
OTL, line 4: Please update mail

openrr-plugin/src/lib.rs
taiki-e: - Without it, `args` will be warned as an unused variable.
- There are two major ways to silence that warning, `#[allow(..)]` and leading `_`, but the former is not nice to look at and the latter is not nice to look when displayed in the documentation. Also, rust-analyzer generates method templates based on the trait definitions, so the user has to start writing by removing the leading `_`.

As a workaround for those problems, `drop` or `let _ =` sometimes be used in the default method for public traits.

- https://docs.rs/syn/1.0.72/src/syn/parse.rs.html#580-583
- https://github.com/serde-rs/serde/blob/1094e2d334f12705957f579f49adfc6118a1de1f/serde/src/ser/mod.rs#L1948
- https://github.com/tokio-rs/valuable/blob/c3efa86b4ac919fc14fe642eb5dfa55916e46e6f/valuable/src/visit.rs#L6

taiki-e: Switched from `drop` to `let _ =` because that seems to be a more widely used pattern for that purpose.

OTL: I see, thanks!

OTL (approved):

PR #310, opened 2021-05-12
Add test for partial joint trajectory client

chatblanc-ciel:
creating test function for partial joint trajectory client

### DIFF
--- a/arci/src/clients/partial_joint_trajectory_client.rs
+++ b/arci/src/clients/partial_joint_trajectory_client.rs
@@ -119,8 +119,8 @@ mod tests {
         last_trajectory: Arc<Mutex<Vec<TrajectoryPoint>>>,
     }
     impl JointTrajectoryClient for DummyFull {
-        fn joint_names(&self) -> &[String] {
-            &self.name
+        fn joint_names(&self) -> Vec<String> {
+            self.name.clone()
         }
 
         fn current_joint_positions(&self) -> Result<Vec<f64>, Error> {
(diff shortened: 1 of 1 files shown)


arci/src/clients/partial_joint_trajectory_client.rs
```diff
@@ -103,3 +103,249 @@ where
...
+            ),
+            TrajectoryPoint::new(
+                vec![3.4_f64, 5.8_f64, 0.1_f64, 2.5_f64],
+                std::time::Duration::from_secs(2),
+            ),
+        ];
+        let correct = vec![3.4_f64, 5.8_f64, 0.1_f64, 2.5_f64];
+
+        let partial = PartialJointTrajectoryClient::new(joint_names.clone(), client.clone());
+        let result = partial.send_joint_trajectory(trajectories);
```
chatblanc-ciel, line 340: if trajectories length is different from client name length,
I think ``fn send_joint_trajectory`` needs assert macro.

Now, part of trajectories are copied and others are zero.
length of the part of trajectories is same as client name length.

chatblanc-ciel (commented):
purpose

a-miyadera:
In actual situation, the length of joint(_name)s of ParitialJointTrajectoryClinet is shorter than that of shared_client (full JointTrajectoryClient).
So, we may need tests in that condition.

arci/src/clients/partial_joint_trajectory_client.rs
```diff
@@ -103,3 +103,410 @@ where
...
+        assert_eq!(
+            format!("{:?}", client),
+            format!("{:?}", partial.shared_client)
+        );
+        assert_eq!(
+            format!("{:?}", partial.full_joint_names),
+            "[\"part1\", \"high\", \"part2\"]"
+        );
+    }
+
```
a-miyadera, line 202: What's the difference between `fn test_fn_copy_joint_position_for_from_joint()` and `fn test_fn_copy_joint_position_for_to_joint()`?

a-miyadera (approved):

PR #317, opened 2021-05-19
Add purpose for fn copy joint name

chatblanc-ciel:


OTL:
@mitsuharu-kojima Can you double-check this?
I think this is correct.

OTL:
Let's change the signature to return Result.

PR #329, opened 2021-05-28
[aric-ros] Use rustros_tf for TransformResolver

mitsuharu-kojima:


OTL (approved):

OTL (approved):

OTL:
rosrust_msgs seems to need ROS environments.
I think we can skip it by arci-ros/build.rs like approach.

https://github.com/openrr/openrr/blob/main/arci-ros/build.rs

OTL (approved):

OTL (approved):

PR #334, opened 2021-06-01
[arci-urdf-viz] Fix leak

taiki-e:
The thread has a cloned `Arc`, so this drop will never be called and `is_dropping` flag will never set to `true`.

https://github.com/openrr/openrr/blob/36a3e59b2f6dcf9c8faa281fbd77783fd6531706/arci-urdf-viz/src/client.rs#L315-L319

And the thread will not terminate unless `is_dropping` flag is `true`.

https://github.com/openrr/openrr/blob/36a3e59b2f6dcf9c8faa281fbd77783fd6531706/arci-urdf-viz/src/client.rs#L244

Therefore, the thread and `Arc<UrdfVizWebClientInner>` will continue to live until the end of the program even if the client is dropped.

### DIFF
--- a/arci-urdf-viz/src/client.rs
+++ b/arci-urdf-viz/src/client.rs
@@ -1,10 +1,7 @@
 use std::{
     collections::HashMap,
     mem,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc, Mutex,
-    },
+    sync::{Arc, Mutex},
     thread::sleep,
     time::Duration,
 };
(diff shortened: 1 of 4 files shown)


arci-urdf-viz/src/client.rs
```diff
@@ -206,60 +217,92 @@ impl UrdfVizWebClient {
...
+            impl Drop for Bomb {
+                fn drop(&mut self) {
+                    self.0
+                        .threads
+                        .lock()
+                        .unwrap()
+                        .has_send_joint_positions_thread = false;
+                    debug!("terminating send_joint_positions_thread");
+                }
+            }
```
taiki-e, line 284: Previously, if a panic occurred in the thread, the thread would terminate, but panic would not be propagated to the main thread, and subsequent `send_joint_positions` calls would never be sent to urdf-viz.

This patch sets the flag to false when the thread terminated so that subsequent `send_joint_positions` calls are panicked if the termination of the thread is due to panic.

OTL (approved):

PR #348, opened 2021-06-04
Improve `fn new`

chatblanc-ciel:
change return ``Result``

test setting

### DIFF
--- a/arci-urdf-viz/src/client.rs
+++ b/arci-urdf-viz/src/client.rs
@@ -83,7 +83,8 @@ fn create_joint_trajectory_clients_inner(
             .into());
         }
         let client =
-            arci::PartialJointTrajectoryClient::new(config.joint_names, all_client.clone());
+            arci::PartialJointTrajectoryClient::try_new(config.joint_names, all_client.clone())
+                .unwrap();
         let client: Arc<dyn JointTrajectoryClient> = if config.wrap_with_joint_velocity_limiter {
             if config.wrap_with_joint_position_limiter {
                 Arc::new(new_joint_position_limiter(
--- a/arci-urdf-viz/src/client.rs
+++ b/arci-urdf-viz/src/client.rs
@@ -83,8 +83,7 @@ fn create_joint_trajectory_clients_inner(
             .into());
         }
         let client =
-            arci::PartialJointTrajectoryClient::try_new(config.joint_names, all_client.clone())
-                .unwrap();
+            arci::PartialJointTrajectoryClient::new(config.joint_names, all_client.clone())?;
         let client: Arc<dyn JointTrajectoryClient> = if config.wrap_with_joint_velocity_limiter {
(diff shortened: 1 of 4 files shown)


arci/src/clients/partial_joint_trajectory_client.rs
```diff
@@ -30,13 +30,33 @@ impl<C> PartialJointTrajectoryClient<C>
...
+            return Err(format!(
+                "Joint name length is illegal. (full = { }, partial = { })",
+                full_joint_names.len(),
+                joint_names.len()
+            ));
+        }
+
+        let mut partial = Vec::with_capacity(joint_names.len());
+        for joint_name in joint_names.iter() {
+            if let Some(_) = full_joint_names.iter().find(|&x| x == joint_name) {
```
a-miyadera, line 45: I think it's enough to return Err at the moment when any item of joint_names is not found in full_joint_names.

chatblanc-ciel: I changed it.
Please, check.

arci-urdf-viz/src/client.rs
```diff
@@ -83,7 +83,8 @@ fn create_joint_trajectory_clients_inner(
             .into());
         }
         let client =
-            arci::PartialJointTrajectoryClient::new(config.joint_names, all_client.clone());
+            arci::PartialJointTrajectoryClient::new(config.joint_names, all_client.clone())
```
chatblanc-ciel, line 86: Config.joint_names is 0 length.
What's happen?

OTL: I think it should be an error.

arci/src/clients/partial_joint_trajectory_client.rs
```diff
@@ -30,13 +30,59 @@ impl<C> PartialJointTrajectoryClient<C>
...
+
+        // check redundant joint name of partial
+        let input_len = joint_names.len();
+        let mut unique_joint_names = joint_names
+            .clone()
+            .into_iter()
+            .collect::<HashSet<String>>()
+            .into_iter()
+            .collect::<Vec<String>>();
+        unique_joint_names.sort();
```
a-miyadera, line 62: Is it necessary to sort?
And, is it necessary to be Vec? (HashSet also has `len` method.)

chatblanc-ciel: I forget to eliminate sort. It is Relics of the past.
Changing Vec is same.
In preview version, unique is necessary for Partial

ros_entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
a-miyadera, line 1: This file is not needed.
Please, delete.

a-miyadera (approved):

PR #344, opened 2021-06-03
[arci-ros2] Allow arci-ros2 to be used as plugin

taiki-e:
~~Depend on #319~~

### Examples

```sh
# build plugins
cargo build -p arci-ros2 --features ros2
cargo build -p openrr-plugin-example

# run gazebo
gazebo --verbose /opt/ros/foxy/share/gazebo_plugins/worlds/gazebo_ros_diff_drive_demo.world
# run openrr_apps_velocity_sender
cargo run -p openrr-apps --bin openrr_apps_velocity_sender -- \
    --config-path ./openrr-apps/config/sample_robot_client_config_for_plugin.toml \
    --config='move_base="arci_ros2";plugins.arci_ros2.instances[0].args="topic=\"/demo/cmd_demo\""'
```

PR #353, opened 2021-06-10
Switch to cargo-llvm-cov

taiki-e:
Switch to [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov).

~~Depend on #344~~

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -2,40 +2,46 @@ name: Code Coverage
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 env:
   CARGO_INCREMENTAL: 0
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
 
-defaults:
-  run:
-    shell: bash -ieo pipefail {0}
-
 jobs:
   coverage:
     runs-on: ubuntu-latest
-    container:
-      image: docker://openrr/openrr-coverage-ci:ros2
-      options: --security-opt seccomp=unconfined
-    env:
-      HOME: /root
     steps:
       - uses: actions/checkout@v2
       - run: ci/ubuntu-install-dependencies.sh
+      - run: |
+          # ros2
+          sudo curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key  -o /usr/share/keyrings/ros-archive-keyring.gpg
+          echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/ros-archive-keyring.gpg] http://packages.ros.org/ros2/ubuntu $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/ros2.list > /dev/null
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yaml
```diff
@@ -2,40 +2,50 @@ name: Code Coverage
...
-      options: --security-opt seccomp=unconfined
-    env:
-      HOME: /root
     steps:
       - uses: actions/checkout@v2
-      - run: ci/ubuntu-install-dependencies.sh
+      - run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            cmake build-essential libudev-dev xorg-dev libglu1-mesa-dev libasound2-dev libclang-dev \
```
OTL, line 22: Is there any reason why avoid using `ci/ubuntu-install-dependencies.sh`?

OTL: And it might be better to be updated.

```
cmake build-essential (openrr-planner (assimp-sys))
libudev-dev (arci-gamepad-gilrs)
xorg-dev libglu1-mesa-dev (openrr-gui (iced))
libasound2-dev (arci-speak-audio)
```

.github/workflows/coverage.yaml
```diff
@@ -2,40 +2,50 @@ name: Code Coverage
...
+            curl gnupg2 lsb-release
+
+          # ros2
+          sudo curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key  -o /usr/share/keyrings/ros-archive-keyring.gpg
+          echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/ros-archive-keyring.gpg] http://packages.ros.org/ros2/ubuntu $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/ros2.list > /dev/null
+          # ros1
+          sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
+          sudo apt-key adv --keyserver 'hkp://keyserver.ubuntu.com:80' --recv-key C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
```
OTL, line 30: At first, I don't want to use docker image.
But do you have any reason why not use docker? Just asking.

I mean ros:foxy-ros1-bridge or something, not openrr-ci.

.github/workflows/coverage.yaml
taiki-e: Good point. That was a change made when I was trying with the ros's docker image, so I think we can probably use `ci/ubuntu-install-dependencies.sh`.

.github/workflows/coverage.yaml
taiki-e: I only tried ros:foxy-ros1-bridge and openrr-ci, but after #344, I could not get the build to work properly with the docker.
There seems to be some package installation missing that is needed for FFI, but I could not found out what it is.

OTL: I see.

OTL (approved):

PR #357, opened 2021-06-14
[arci-ros2] Implement ROS2 navigation2 using r2r actions branch

OTL:

PR #360, opened 2021-06-16
[arci-ros2] Store Node globally

taiki-e:
`r2r::Node` is now `Send`. https://github.com/sequenceplanner/r2r/commit/5ab5ab5e3f5f0f137886fd98654b6e7cc99ea3b1


OTL (approved):
It's OK, but my change might change everything in near future.

https://github.com/openrr/openrr/pull/357

PR #368, opened 2021-06-23
[arci-urdf-viz] Remove web_server.rs

taiki-e:
We have switched to cargo-llvm-cov, so this should be no longer necessary.

Depend on https://github.com/openrr/urdf-viz/pull/110.


OTL:
@taiki-e I want to (update &) merge this if possible.

OTL (approved):

PR #374, opened 2021-06-28
[CI] Install libxkbcommon0

mitsuharu-kojima:


OTL (approved):
I understand the situation, thank you for your quick fix!
iced has been changed?

PR #385, opened 2021-07-07
[openrr-client] Correct variable names: using_joints->used_joints

ssr-yuki:
`used_joints` is more suitable than `using_joints`


ssr-yuki:
Since this renaming should be discussed more deeply, I temporarily close it.

PR #395, opened 2021-07-14
[arci-speak-audio] Fix an issue where audio stop

taiki-e:


OTL (approved):

taiki-e:
Updated to catch as many errors as possible before thread is spawned.

OTL (approved):

PR #397, opened 2021-07-14
[openrr-client] Use SelfCollisionChecker in collision_avoidance_client

ssr-yuki:


OTL (commented):
It seems there are no reasons to use SelfCollisionChecker for this client.

ssr-yuki:
You're right. I opened this by my misunderstanding. Sorry.

PR #403, opened 2021-07-21
[arci] `move base` purpose

chatblanc-ciel:


arci/src/traits/move_base.rs
```diff
@@ -32,12 +32,16 @@ impl BaseVelocity {
 impl std::ops::Mul<f64> for BaseVelocity {
     type Output = Self;
 
-    fn mul(self, rhs: f64) -> Self::Output {
```
chatblanc-ciel, line 35: `Self{ }` is generate new instance.
But this function need self. I think it is waste action.

arci/src/traits/move_base.rs
```diff
@@ -32,12 +32,16 @@ impl BaseVelocity {
...
-        Self {
+    fn mul(mut self, rhs: f64) -> Self::Output {
+        /*     Self {
             x: self.x * rhs,
```
chatblanc-ciel, line 37: need delete spare.

chatblanc-ciel (commented):
request

taiki-e (approved):

PR #404, opened 2021-07-21
[openrr-teleop] Add test for `move_base`

chatblanc-ciel:
Add test function

- `fn get`(function name is not same. )

- `fn proc`
four type variation test

### DIFF
--- a/openrr-teleop/Cargo.toml
+++ b/openrr-teleop/Cargo.toml
@@ -28,4 +28,5 @@ tokio = { version = "1.0", features = ["rt-multi-thread", "sync", "time"] }
 tracing = { version = "0.1", features = ["log"] }
 
 [dev-dependencies]
+approx = "0.5.0"
 tokio = { version = "1.0", features = ["full"] }
--- a/openrr-teleop/Cargo.toml
+++ b/openrr-teleop/Cargo.toml
@@ -28,5 +28,5 @@ tokio = { version = "1.0", features = ["rt-multi-thread", "sync", "time"] }
 tracing = { version = "0.1", features = ["log"] }
 
 [dev-dependencies]
-approx = "0.5.0"
+assert_approx_eq = "1.1.0"
 tokio = { version = "1.0", features = ["full"] }
--- a/openrr-teleop/src/move_base.rs
+++ b/openrr-teleop/src/move_base.rs
@@ -116,4 +116,34 @@ mod tests {
         assert!(!node.is_enabled);
         assert!(!node.is_turbo);
     }
+
+    #[test]
+    fn test_move_node_get() {
+        let mode = String::from("tested");
+        let base = DummyMoveBase::new();
+        let node = MoveBaseNode::new(mode.clone(), base);
+
+        let base_mode = node.mode();
+        assert_eq!(base_mode, mode);
+        let base_sub_mode = node.submode();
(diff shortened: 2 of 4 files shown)


openrr-teleop/src/move_base.rs
```diff
@@ -116,4 +116,76 @@ mod tests {
...
+            submode: "".to_string(),
+            vel: BaseVelocity::default(),
+            is_enabled: false,
+            is_turbo: false,
+        };
+        node.proc().await;
+        assert_eq!(
+            format!("{:?}", node.vel),
+            format!("{:?}", BaseVelocity::default())
+        );
```
a-miyadera, line 147: `node.vel` will not be updated regardless of the values of `is_enbled` and `is_turbo`.
I suppose it's better to set node.vel to non-zero first and compare with return value of `node.move_base.current_velocity()` after calling `proc()`.
And, when comparing, it might be better to compare each field with `assert_approx_eq` macro rather than to compare formatted strings.
The same below.

chatblanc-ciel: - Using `assert_approx_eq`
It is ok. I'll apply.

- `node.vel` will not be updated
I think if `is_turbo` is true, `proc` works `vel` twice and others can't affect.
So, I tested these boolean pattern.

openrr-teleop/Cargo.toml
```diff
@@ -28,4 +28,5 @@ tokio = { version = "1.0", features = ["rt-multi-thread", "sync", "time"] }
 tracing = { version = "0.1", features = ["log"] }
 
 [dev-dependencies]
+approx = "0.5.0"
```
OTL, line 31: If you have no reason to use approx instead of assert_approx_eq, please use assert_approx_eq to keep consistency and not to increase dependencies.

openrr-teleop/src/move_base.rs
```diff
@@ -116,4 +117,95 @@ mod tests {
...
+        let base_sub_mode = node.submode();
+        assert_eq!(base_sub_mode, &String::from(""));
+    }
+
+    #[tokio::test]
+    async fn test_move_node_proc() {
+        let mode = String::from("tested");
+        let x = 1.2_f64;
+        let y = 3.5;
+        let theta = 1.8;
```
OTL, line 138: Why added _f64 only for `x`?
Can't compile without it?

OTL: And you can use const if the values are not changed.

```rust
const x: f64 X = 1.2;
...
```

OTL: And it's better to add some information about what x means.
(reference, target, initial, or something)
I mean `const SOME_INFORMATION_X: f64 = 1.2;` is better.

chatblanc-ciel: Other valiables are forgot to add it.

But, no explicit type is judged `float` probably.
I think compiler miss to read true sentences in test function, it is not test.
Using implicit delaration is good for reading when you use it in practical scene.
I use usually post-modification, now I heard your opnion and I'm going to use pre-modification if I notice it.

openrr-teleop/Cargo.toml
chatblanc-ciel: ok.

OTL (approved):

PR #407, opened 2021-07-21
[openrr-client] Add a CollisionCheckClient test

ssr-yuki:


openrr-client/src/clients/collision_check_client.rs
```diff
@@ -81,3 +81,47 @@ pub fn create_collision_check_client<P: AsRef<Path>>(
         )),
     )
 }
+
+#[cfg(test)]
```
OTL, line 85: I want you to move the test code into `mod test` because you'll increase more tests. If the test is only one, it has no meaning actually.

```rust
mod test {
```

ssr-yuki: You are right, I or another developer would add tests in the future.

ssr-yuki:
Sorry, I forget to push a sample URDF file copied from openrr-planner.
The CI failure comes from this.

OTL (approved):

PR #409, opened 2021-07-26
Fix minor warning

chatblanc-ciel:
- [x] filename collision
- [x] Clippy warnings


OTL (approved):
Thanks!
I think your change about the `clippy` is right, but do you know why our clippy GitHub actions detect it?

chatblanc-ciel:
I'm good at `clippy`.
But, defference is ` -- -D warnings`, `+nightly`, `set profile minimal`(My profile is `complete` thus official says complete is for language developers).
These occur missing lints, I think there is a possibility.

PR #410, opened 2021-07-27
[arci-ros] Create test for new_joint_position_limiter

chatblanc-ciel:
-  for `fn new_joint_position_limiter`

### DIFF
--- a/arci-ros/src/ros_control_client.rs
+++ b/arci-ros/src/ros_control_client.rs
@@ -493,10 +493,7 @@ mod tests {
     fn test_new_joint_position_limiter() {
         use arci::{DummyJointTrajectoryClient, JointPositionLimit, JointPositionLimiter};
 
-        let client = DummyJointTrajectoryClient::new(vec![
-            String::from("a".to_owned()),
-            String::from("b".to_owned()),
-        ]);
+        let client = DummyJointTrajectoryClient::new(vec!["a".to_owned(), "b".to_owned()]);
         let limits = vec![
             JointPositionLimit::new(-5.0, 5.0),
             JointPositionLimit::new(-5.0, 5.0),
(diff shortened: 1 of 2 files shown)


arci-ros/src/ros_control_client.rs
```diff
@@ -484,3 +484,47 @@ impl SetCompleteCondition for RosControlClient {
...
+
+        let result = new_joint_position_limiter(client, Some(limits.clone()), None);
+        assert!(result.is_ok());
+        let result = result.unwrap();
+
+        let client = DummyJointTrajectoryClient::new(vec![
+            String::from("a".to_owned()),
+            String::from("b".to_owned()),
+        ]);
+        let correct = JointPositionLimiter::new(client, limits);
```
chatblanc-ciel, line 513: How check private field?

chatblanc-ciel (commented):
Help me.

chatblanc-ciel:
Clippy of `GitHub Action` says `"".to_owned()` is error?
I decided these are changed to `String::from()`.

OTL:
> Clippy of `GitHub Action` says `"".to_owned()` is error?
> I decided these are changed to `String::from()`.

Please add the URL when you want to mention about the result of github actions.
I think you're talking about below.
https://github.com/openrr/openrr/runs/3258249000?check_suite_focus=true

It says that do not convert String into String. It does not say "do not use to_owned()".
Please revert `String::from` part.

chatblanc-ciel:
> > Clippy of `GitHub Action` says `"".to_owned()` is error?
> > I decided these are changed to `String::from()`.
> 
> Please add the URL when you want to mention about the result of github actions.
> I think you're talking about below.
> https://github.com/openrr/openrr/runs/3258249000?check_suite_focus=true
> 
> It says that do not convert String into String. It does not say "do not use to_owned()".
> Please revert `String::from` part.

Yes, I get it.　Thanks.

OTL (approved):

PR #420, opened 2021-08-04
[openrr-planner] Get collision link names before resetting

ssr-yuki:
In current implementation, both `plan()` and `plan_avoid_self_collision()` functions never show which links will collide, since resetting joint angles runs before identify the link names.

This PR solves this problem just by inverting algorithms: first, get collision link names with reference joint angles; then, reset joint angles.


OTL (approved):
OMG! Thanks!

PR #421, opened 2021-08-04
[planner & client] Correct DoF in collision check tests

ssr-yuki:
A robot defined in `sample.urdf` has only 8 movable joints. This PR corrects mistakes I made in #396 and #407  .


OTL (approved):
OMG! Thanks

PR #423, opened 2021-08-05
[openrr-remote] Add openrr-remote

taiki-e:


openrr-remote/src/lib.rs
```diff
@@ -0,0 +1,735 @@
...
+        Ok(tonic::Response::new(()))
+    }
+
+    async fn cancel(&self, _: tonic::Request<()>) -> Result<tonic::Response<()>, tonic::Status> {
+        arci::Navigation::cancel(&self.inner).map_err(|e| tonic::Status::unknown(e.to_string()))?;
+        Ok(tonic::Response::new(()))
+    }
+}
+
+// =============================================================================
```
OTL, line 376: I think it's better to divide into multiple files if you need this ================

openrr-remote/src/lib.rs
```diff
@@ -0,0 +1,735 @@
...
+            y: val.y,
+            theta: val.theta,
+        }
+    }
+}
+
+impl From<arci::Isometry2<f64>> for pb::Isometry2 {
+    fn from(val: arci::Isometry2<f64>) -> Self {
+        Self {
+            rotation: Some(pb::UnitComplex {
```
OTL, line 559: why we need Option?

openrr-remote/src/lib.rs
```diff
@@ -0,0 +1,735 @@
...
+
+impl From<pb::Isometry2> for arci::Isometry2<f64> {
+    fn from(val: pb::Isometry2) -> Self {
+        let translation = val.translation.unwrap();
+        let rotation = val.rotation.unwrap();
+        Self::from_parts(
+            nalgebra::Translation2::new(translation.x, translation.y),
+            nalgebra::UnitComplex::from_complex(nalgebra::Complex {
+                re: rotation.re,
+                im: rotation.im,
```
OTL, line 579: why you chose complex for rotation?

openrr-remote/src/lib.rs
taiki-e: In proto3, all fields are optional: https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md

openrr-remote/src/lib.rs
taiki-e: I copied them from openrr-plugin. If there is another good way to decode/encode Isometry2, suggestions are welcome.

openrr-remote/src/lib.rs
OTL: I see. I know that.

openrr-remote/src/lib.rs
OTL: I thought that just an angle (float) is enough (without much thought).

OTL (approved):
Anyway i approve

openrr-remote/src/lib.rs
taiki-e: (We won't need to stick to the file layout issue here since we will eventually do something like https://github.com/openrr/openrr/pull/431.)

PR #425, opened 2021-08-06
[openrr-teleop] Add test for `control node`

chatblanc-ciel:
- [x] change returning `Result`
- [x] add test for `ok`
- [x] add test for `error`

### DIFF
--- a/arci-gamepad-keyboard/examples/keyboard_check_keymap.rs
+++ b/arci-gamepad-keyboard/examples/keyboard_check_keymap.rs
@@ -1,6 +1,9 @@
+#[cfg(unix)]
 use arci::gamepad::{Gamepad, GamepadEvent};
+#[cfg(unix)]
 use arci_gamepad_keyboard::*;
 
+#[cfg(unix)]
 #[tokio::main]
 async fn main() {
     tracing_subscriber::fmt::init();
@@ -15,3 +18,6 @@ async fn main() {
         }
     }
 }
+
+#[cfg(windows)]
+fn main() {}
--- a/openrr-teleop/src/control_nodes_config.rs
+++ b/openrr-teleop/src/control_nodes_config.rs
@@ -118,3 +118,27 @@ impl ControlNodesConfig {
         nodes
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_gen() {
+        // JoyJointTeleopConfig
+        let config = JoyJointTeleopNodeConfig{
+            mode :String::from("a"),
+            joint_step :1.0_f64,
+            step_duration_secs :0.001_f64
+        };
+        let _joy_tconfig = JoyJointTeleopConfig{
+            client_name :String::from("b"),
+            config
+        };
+
+        // IkNodeTeleopConfig
+        let _iknode_tconfig = IkNodeTeleopConfig{
+
+        };
+    }
+}
--- a/openrr-teleop/src/control_nodes_config.rs
+++ b/openrr-teleop/src/control_nodes_config.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashMap, path::PathBuf, sync::Arc};
 
 use arci::{JointTrajectoryClient, MoveBase, Speaker};
-use openrr_client::{ArcRobotClient, IkSolverWithChain, JointsPose};
+use openrr_client::{ArcRobotClient, Error, IkSolverWithChain, JointsPose};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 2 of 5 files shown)


openrr-teleop/src/control_nodes_config.rs
```diff
@@ -115,6 +150,192 @@ impl ControlNodesConfig {
...
+            step_duration_secs: default_step_duration_secs(),
+        };
+
+        assert_approx_eq!(config.move_step_angular[0], 0.05);          // Config default value check(move_step_angular)
+        assert_approx_eq!(config.move_step_angular[1], 0.05);
+        assert_approx_eq!(config.move_step_angular[2], 0.17);
+        assert_approx_eq!(config.move_step_linear[0], 0.01);            // Config default value check(move_step_linear)
+        assert_approx_eq!(config.move_step_linear[1], 0.01);
+        assert_approx_eq!(config.move_step_linear[2], 0.01);
+        assert_approx_eq!(config.step_duration_secs, 0.1);              // Config default value check(step_duration_secs)
```
a-miyadera, line 200: Are these needed?
We are not testing `IkNodeConfig` here.

openrr-teleop/src/control_nodes_config.rs
```diff
@@ -115,6 +150,192 @@ impl ControlNodesConfig {
...
+        let nodes = ctrl_node_config.create_control_nodes(
+            None,
+            Arc::new(robot_client),
+            speaker,
+            &HashMap::new(),
+            &HashMap::new(),
+            Some(Arc::new(DummyMoveBase::default())),
+            Vec::new(),
+        );
+        assert!(nodes.is_err());
```
a-miyadera, line 262: I want to know why this must be an error.
Could you write in a comment?

openrr-teleop/src/control_nodes_config.rs
chatblanc-ciel: I'm sorry, this block is written first so these have not removed yet.
I think removing it is good.

openrr-teleop/src/control_nodes_config.rs
```diff
@@ -115,6 +150,228 @@ impl ControlNodesConfig {
...
+        };
+        let ctrl_node_config = ControlNodesConfig {
+            move_base_mode: Some(String::from("a")),
+            joy_joint_teleop_configs: vec![jj_config],
+            ik_node_teleop_configs: vec![ik_config],
+            joints_pose_sender_config: None,
+            command_configs: vec![rc_config],
+        };
+
+        // Check create control node(Error:HashMap length)
```
a-miyadera, line 232: Is this right?
It seems to be a matter of whether the value for the specified key exists in the HashMap or not, rather than the length.

openrr-teleop/src/control_nodes_config.rs
```diff
@@ -115,6 +150,228 @@ impl ControlNodesConfig {
...
+            &HashMap::new(),
+            &HashMap::new(),
+            Some(Arc::new(DummyMoveBase::default())),
+            Vec::new(),
+        );
+        assert!(nodes.is_err());
+        let err = nodes.err().unwrap();
+        println!("{}", err);
+
+        // Check create control node(Error:name mismatch)
```
a-miyadera, line 257: Explain in more detail.
I want to know which parameter caused the error. (The same applies above.)

openrr-teleop/src/control_nodes_config.rs
chatblanc-ciel: I'm sorry that I think length 0 same as not contained key.
But `specified key exists in the HashMap or not` is more correct.

openrr-teleop/src/control_nodes_config.rs
```diff
@@ -115,6 +154,230 @@ impl ControlNodesConfig {
...
+            OpenrrClientsConfig::default(),
+            HashMap::new(),
+            HashMap::new(),
+            None,
+            None,
+            None,
+        );
+        assert!(robot_client.is_ok());
+        let robot_client = robot_client.unwrap();
+        let nodes = ctrl_node_config.create_control_nodes(
```
a-miyadera, line 249:
```suggestion
        // the 4th argument `joint_trajectory_client_map` had to have keys same with the items in `joy_joint_teleop_configs`
        let nodes = ctrl_node_config.create_control_nodes(
```

openrr-teleop/src/control_nodes_config.rs
```diff
@@ -115,6 +154,230 @@ impl ControlNodesConfig {
...
+    use crate::{default_move_step_angular, default_move_step_linear, default_step_duration_secs};
+
+    #[test]
+    fn test_gen() {
+        // JoyJointTeleopConfig
+        let config = JoyJointTeleopNodeConfig {
+            mode: String::from("a"),
+            joint_step: 1.0_f64,
+            step_duration_secs: default_step_duration_secs(),
+        };
```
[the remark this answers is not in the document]
chatblanc-ciel: I  agree but why I  can use `toml style`？

These default function have  what purpose？I think using public(`pub(crate)` is wrong, I  want to  write `pub`. Sorry.).

openrr-teleop/src/ik.rs
```diff
@@ -242,15 +242,15 @@ pub struct IkNodeConfig {
     pub step_duration_secs: f64,
 }
 
-const fn default_move_step_angular() -> [f64; 3] {
+pub(crate) const fn default_move_step_angular() -> [f64; 3] {
```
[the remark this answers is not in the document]
chatblanc-ciel: Yes, it is.
But, I want write `pub` because I think this function is using public.

openrr-teleop/src/control_nodes_config.rs
chatblanc-ciel: Thank you for advice.
I'll change it.

openrr-teleop/src/control_nodes_config.rs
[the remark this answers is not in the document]
a-miyadera: You can use toml style because `IkNodeConfig` struct implements `Deserialize` trait.
https://github.com/openrr/openrr/blob/33a1ba24b925b5a79f382c011bd236c169e8b1fe/openrr-teleop/src/ik.rs#L233
(I have shown it as an example, but do not recommend it.
You can use any values for the tests.)

Those default functions are only for providing default values on deserializing.
(cf. https://serde.rs/attr-default.html)
And they are not intended to be used publicly.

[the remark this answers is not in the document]
chatblanc-ciel: Thank you for explanation.
This problem solve by to change to using specific value and returning private function.

arci-gamepad-keyboard/examples/keyboard_check_keymap.rs
```diff
@@ -1,6 +1,9 @@
+#[cfg(unix)]
 use arci::gamepad::{Gamepad, GamepadEvent};
+#[cfg(unix)]
 use arci_gamepad_keyboard::*;
 
+#[cfg(unix)]
```
a-miyadera, line 6: Is this not buildable on Windows?

a-miyadera (approved):

OTL (approved):

PR #408, opened 2021-07-21
Support urdf-viz client without config

taiki-e:
Depend on https://github.com/openrr/urdf-viz/pull/96.

```sh
cargo run --manifest-path ../urdf-viz/Cargo.toml -- ./openrr-planner/sample.urdf &

# openrr_apps_joint_position_sender
cargo run -p openrr-apps --bin openrr_apps_joint_position_sender

# openrr_apps_robot_teleop
cargo run -p openrr-apps --bin openrr_apps_robot_teleop
```

### TODO

- [x] polish & merge https://github.com/openrr/urdf-viz/pull/96
- [x] openrr_apps_joint_position_sender
- [x] openrr_apps_velocity_sender
- [x] openrr_apps_robot_command
- [x] openrr_apps_robot_teleop ~~(Note: teleop config is still needed)~~
- [x] support overwriting default config even if config-path is not specified

PR #427, opened 2021-08-12
[openrr-apps] Clean up config-related code

taiki-e:
Addresses https://github.com/openrr/openrr/pull/408#pullrequestreview-719886603


OTL (approved):
Thanks!

PR #430, opened 2021-08-13
[openrr-teleop] Add test for `default function`

chatblanc-ciel:
- ik.rs
- joints.rs

### DIFF
--- a/openrr-teleop/src/ik.rs
+++ b/openrr-teleop/src/ik.rs
@@ -220,14 +220,48 @@ pub struct IkNodeConfig {
     pub step_duration_secs: f64,
 }
 
-fn default_move_step_angular() -> [f64; 3] {
+const fn default_move_step_angular() -> [f64; 3] {
     [0.05, 0.05, 0.17]
 }
 
-fn default_move_step_linear() -> [f64; 3] {
+const fn default_move_step_linear() -> [f64; 3] {
     [0.01, 0.01, 0.01]
 }
 
-fn default_step_duration_secs() -> f64 {
+const fn default_step_duration_secs() -> f64 {
     0.1
+}
+
+#[cfg(test)]
+mod tests {
+    use assert_approx_eq::*;
+
+    use super::*;
+
+    #[test]
+    fn test_default_move_step_angular() {
+        let def = default_move_step_angular();
+
+        assert_eq!(def.len(), 3_usize);
+        assert_approx_eq!(def[0], 0.05_f64);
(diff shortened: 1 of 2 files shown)


openrr-teleop/src/ik.rs
```diff
@@ -226,3 +226,39 @@ fn default_move_step_linear() -> [f64; 3] {
...
+
+    #[test]
+    fn test_default_move_step_angular() {
+        let def = default_move_step_angular();
+        let mut def_iter = def.iter();
+
+        assert_approx_eq!(def_iter.next().unwrap(), 0.05_f64);
+        assert_approx_eq!(def_iter.next().unwrap(), 0.05_f64);
+        assert_approx_eq!(def_iter.next().unwrap(), 0.17_f64);
+        assert_eq!(def_iter.next(), None);
```
a-miyadera, line 244: How about using elements of the array directory (e.g. `def[0]`) instead of getting them via iterator?
I think it would become simpler and be easier to convey your intentions.

chatblanc-ciel: I'm sorry, you are right very much.
I do so because I felt like it. And no reason.

I'm going to redo at next time.

openrr-teleop/src/ik.rs
```diff
@@ -231,3 +231,37 @@ fn default_move_step_linear() -> [f64; 3] {
...
+    use super::*;
+
+    #[test]
+    fn test_default_move_step_angular() {
+        let def = default_move_step_angular();
+
+        assert_approx_eq!(def.get(0).unwrap(), 0.05_f64);
+        assert_approx_eq!(def.get(1).unwrap(), 0.05_f64);
+        assert_approx_eq!(def.get(2).unwrap(), 0.17_f64);
+        assert_eq!(def.get(3), None);
```
a-miyadera, line 248: I prefer to use index access(i.e. def[0], def[1], def[2]), and to confirm length of the array.
But it's OK if it works.

a-miyadera (approved):

openrr-teleop/src/ik.rs
chatblanc-ciel: I think using `.get()` everytime is bad choice.
However, using it is good if length is not obvious.

In the test, I want to check including length.

chatblanc-ciel: P.S.
This code is working and `def.get(*)` is returned `Some(def[*])` if elements exist.

a-miyadera (approved):

PR #431, opened 2021-08-16
[openrr-plugin] Generate implementations from trait definitions

taiki-e:
Except for Gamepad and JointTrajectoryClient.


tools/gen.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
OTL, line 1: I think it's better if the file name were more descriptional.
I mean rename to `gen-xxxx.sh`.
I guess that you are thinking about adding more functionality in this script, so how about `gen-code.sh` or something.

OTL (commented):
Sorry for late review

tools/gen.sh
taiki-e: renamed to `gen-code.sh`

OTL (approved):

PR #432, opened 2021-08-18
Update cargo-llvm-cov

taiki-e:
Released the first non-alpha version of cargo-llvm-cov: https://github.com/taiki-e/cargo-llvm-cov/releases/tag/v0.1.0


OTL (approved):
Congratulations!

PR #433, opened 2021-08-18
[planner] Use generics in SelfCollisionChecker refering to JointPathPlanner

ssr-yuki:
SelfCollisionChecker is similar to JointPathPlaner, and we start to create a new struct describing common parameters of them.

As the first step of the challenge, this PR introduces generics in SelfCollisionChecker as used in JointPathPlanner.

Reference:
https://github.com/openrr/openrr/blob/52af54eaf69ad0b9152dd2d1e6000dd36c7e85a9/openrr-planner/src/planner/joint_path_planner.rs#L25-L390

### DIFF
--- a/openrr-planner/src/collision/self_collision_checker.rs
+++ b/openrr-planner/src/collision/self_collision_checker.rs
@@ -61,11 +61,11 @@ where
         positions: &[N],
         duration: std::time::Duration,
     ) -> Result<()> {
+        let duration_f64 = num_traits::NumCast::from::<f64>(duration.as_secs_f64()).unwrap();
         match interpolate(
             &[current.to_vec(), positions.to_vec()],
-            num_traits::NumCast::from::<f64>(duration.as_secs_f64()).unwrap(),
-            self.time_interpolate_rate
-                .mul(num_traits::NumCast::from::<f64>(duration.as_secs_f64()).unwrap()),
+            duration_f64,
+            self.time_interpolate_rate.mul(duration_f64),
         ) {
             Some(interpolated) => {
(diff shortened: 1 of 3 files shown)


openrr-planner/src/collision/self_collision_checker.rs
```diff
@@ -48,14 +57,15 @@ impl SelfCollisionChecker {
...
-        positions: &[f64],
+        current: &[N],
+        positions: &[N],
         duration: std::time::Duration,
     ) -> Result<()> {
         match interpolate(
             &[current.to_vec(), positions.to_vec()],
-            duration.as_secs_f64(),
-            duration.as_secs_f64() * self.time_interpolate_rate,
+            num_traits::NumCast::from::<f64>(duration.as_secs_f64()).unwrap(),
```
OTL, line 66: How about using variable to reduce the twice complicated conversion.

```rust
let duration_f64: f64 = num_traits::NumCast::from::<f64>(duration.as_secs_f64()).unwrap();
```

And aren't there any more easy conversions?

OTL (commented):
Have you checked that this change does not break our closed source?

openrr-planner/src/collision/self_collision_checker.rs
ssr-yuki: Thanks, I accept your suggestion introducing a new variable.
Unfortunately, I do not know nor find any better conversion.

ssr-yuki:
> Have you checked that this change does not break our closed source?

Thank you for reminding me this. Now I open another PR corresponding to this change there.

OTL (approved):

PR #438, opened 2021-08-24
[openrr-apps, aric-ros] Add tracing utility function

mitsuharu-kojima:


openrr-apps/src/utils.rs
```diff
@@ -126,7 +137,119 @@ pub fn init_with_anonymize(name: &str, config: &RobotConfig) {
...
+    #[serde(default = "default_log_prefix")]
+    pub prefix: String,
+}
+
+fn default_log_prefix() -> String {
+    "log".to_string()
+}
+
+fn default_log_directory() -> String {
+    "/tmp".to_string()
```
OTL, line 254: I think `/tmp` does not work on windows.

How about using rust function?
https://doc.rust-lang.org/std/env/fn.temp_dir.html

openrr-apps/src/utils.rs
```diff
@@ -126,7 +137,119 @@ pub fn init_with_anonymize(name: &str, config: &RobotConfig) {
...
+}
+
+#[derive(Deserialize)]
+pub enum LogRotation {
+    MINUTELY,
+    HOURLY,
+    DAILY,
+    NEVER,
+}
+impl Default for LogRotation {
```
OTL, line 230: Add blank line?
See L217, also.

openrr-apps/src/utils.rs
```diff
@@ -126,7 +137,119 @@ pub fn init_with_anonymize(name: &str, config: &RobotConfig) {
 
 pub fn init_tracing() {
     tracing_subscriber::fmt()
-        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .with_env_filter(EnvFilter::from_default_env())
         .with_writer(io::stderr)
         .init();
 }
+
```
OTL, line 144: Could you add some simple tests for these?
(Simple means that you don't have to think of coverage rate.)
The tests will be run on windows too.

OTL (approved):
Thanks!

PR #452, opened 2021-09-16
[arci-ros] add test helper

chatblanc-ciel:
for inspect subscriber message


arci-ros/tests/test_rosrust_utils.rs
```diff
@@ -77,3 +77,41 @@ fn test_convert_system_time_to_ros_time() {
         ALLOWABLE_ERROR
     );
 }
+
+#[test]
+fn test_sub_helper() {
```
a-miyadera, line 82: What's the purpose of this test?
Is it a test of the `subscribe_helper` function?

arci-ros/tests/util/msg_helper.rs
```diff
@@ -0,0 +1,27 @@
...
+/// # subscribe ROS message helper
+///
+/// using for inspect specific massage type.
+/// Message is displayed on screen and sent to ``mpsc receiver``
+///
+/// # Panic!
+///
+/// If subscriber can't be construct, this function is panic.
+///
+#[allow(dead_code)]
```
a-miyadera, line 14: Is this needed?

arci-ros/tests/util/msg_helper.rs
```diff
@@ -0,0 +1,27 @@
...
+pub(crate) fn subscribe_helper<T: rosrust::Message>(
```
a-miyadera, line 15: I'm afraid the name `subscribe_helper` does not indicate the role of this function properly.
How about `create_test_subscriber`, `create_subscriber_for_test` or something.

arci-ros/tests/util/msg_helper.rs
chatblanc-ciel: If this code is not, cllipy says warning.
But be not called is also exist because using test helper.

arci-ros/tests/util/msg_helper.rs
chatblanc-ciel: Ok.

arci-ros/tests/test_rosrust_utils.rs
chatblanc-ciel: Your thinking is right.
I want to check working what I designed first.

a-miyadera: So, this is a test of the test code.
There might be room to discuss whether the test code needs to be tested or not.
Anyway, this file must contain tests only for `rosrust_utils.rs`.
If you want to test the `sucscribe_helper` function, you can write test code in the same file the function is written in (i.e. `msg_helper.rs`).

arci-ros/tests/util/msg_helper.rs
a-miyadera: Certainly.
However, if you write the test code of this function in this file, this attribute might not be required.

a-miyadera (approved):

PR #455, opened 2021-09-22
[openrr-planner] Remove transform update in CollisionDetector

ssr-yuki:
Since a robot model is just a argument, CollisionDetector should not add any changes to it. 
Instead, the same update before collision calculation is implemented in RobotCollisionDetector:

https://github.com/openrr/openrr/blob/d6971f240cb9fc5fd3a84da9a43dfbe8c222be8b/openrr-planner/src/collision/robot_collision_detector.rs#L44-L68

### DIFF
--- a/openrr-planner/src/collision/collision_detector.rs
+++ b/openrr-planner/src/collision/collision_detector.rs
@@ -276,7 +276,6 @@ where
         target_shape: &'a dyn Shape<T>,
         target_pose: &'a na::Isometry3<T>,
     ) -> EnvCollisionNames<'a, 'a, T> {
-        robot.update_transforms();
         EnvCollisionNames::new(self, robot, target_shape, target_pose)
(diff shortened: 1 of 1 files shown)


mitsuharu-kojima (approved):

ssr-yuki:
This should be merged at the same time with a PR embeding RobotCollisionDetector in SelfCollisionChecker, in order to prevent any collision checking without updating robot poses.

ssr-yuki:
After both #457 and #458 are merged, this change does not affect any functions.

OTL:
As we discussed offline, this is a public interface, so there is no guarantee that the transform is updated.
If you want to do this, we need a more complicated `transform()` method which has a `dirty` flag when the parent joint is changed, and update it when `transform()` is called in the `k` library.
Let me close this.

PR #494, opened 2021-10-29
Revert "[arci-speak-audio] pin alsa to 0.5.0"

taiki-e:
This reverts commit 89bf3ad3e7e2ee5c710e01fe4ce86831dc562359.

alsa 0.5.1 has been yanked, so we no longer need to pin it.


OTL (approved):
GJ!

PR #507, opened 2021-11-10
[arci] Add JointPositionDifferenceLimiter

mitsuharu-kojima:


arci-ros/src/ros_control/joint_trajectory_client_wrapper_config.rs
```diff
@@ -77,9 +100,30 @@ pub(crate) fn wrap_joint_trajectory_client(
 ) -> Result<Arc<dyn JointTrajectoryClient>, arci::Error> {
     Ok(if config.wrap_with_joint_velocity_limiter {
         if config.wrap_with_joint_position_limiter {
-            Arc::new(new_joint_position_limiter(
-                new_joint_velocity_limiter(client, config.joint_velocity_limits, urdf_robot)?,
-                config.joint_position_limits,
+            if config.wrap_with_joint_position_difference_limiter {
+                let client = new_joint_position_difference_limiter(
+                    client,
```
OTL, line 105: This initialization sequence (the whole of this function) seems too complicated. I understand that this is caused by the difference of the types, and is not solved easily. But I feel that we have to solve this soon.
(My idea is to use macros to do this, or wrap with Box)

OTL (approved):

PR #532, opened 2021-11-25
[arci] Add `send_goal_pose_with_thresholds` function to `Navigation` trait

a-miyadera:
for navigation in which movement accuracy is not needed

### DIFF
--- a/arci/src/traits/navigation.rs
+++ b/arci/src/traits/navigation.rs
@@ -11,5 +11,16 @@ pub trait Navigation: Send + Sync {
         timeout: std::time::Duration,
     ) -> Result<WaitFuture, Error>;
 
+    #[allow(unused_variables)]
+    fn send_goal_pose_with_thresholds(
+        &self,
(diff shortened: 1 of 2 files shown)


OTL:
https://github.com/openrr/openrr/runs/4323697198?check_suite_focus=true
is failing. I think we need to discuss the interface a little more.

arci/src/traits/navigation.rs
```diff
@@ -11,5 +11,18 @@ pub trait Navigation: Send + Sync {
         timeout: std::time::Duration,
     ) -> Result<WaitFuture, Error>;
 
+    #[allow(unused_variables)]
+    fn send_goal_pose_with_thresholds(
+        &self,
+        goal: Isometry2<f64>,
+        thresholds: [f64; 3],
```
OTL, line 18: Add some comments about what the `thresholds` means, or 
change the name to goal_torrelance_x_y_theta?

PR #540, opened 2021-12-03
Support evaluation of config string in openrr-apps

mitsuharu-kojima:


OTL:
I thought it was okay to always use evaluate=true, but are there any cases where it's not okay?
I think there is, but I can't think of any specific case... If there are not too many, the default true might be fine.

mitsuharu-kojima:
> I thought it was okay to always use evaluate=true, but are there any cases where it's not okay? I think there is, but I can't think of any specific case... If there are not too many, the default true might be fine.

Done

OTL (approved):

PR #570, opened 2022-01-20
[openrr-apps] Warn TOML config that is not valid as TOML

taiki-e:
output example:
```
2022-01-20T04:15:25.912376Z  WARN openrr_apps::utils: config ./openrr-apps/config/sample_robot_client_config_for_urdf_viz.toml is not valid TOML: unexpected character found: `$` at line 44 column 1
```


mitsuharu-kojima:
+1

OTL (approved):

PR #577, opened 2022-02-07
Use lld in Linux CI

taiki-e:


ci/ubuntu-setup-lld.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
+
+LLVM_VERSION=13
+
+codename="$(grep '^VERSION_CODENAME=' /etc/os-release | sed 's/^VERSION_CODENAME=//')"
+echo "deb http://apt.llvm.org/${codename}/ llvm-toolchain-${codename}-13 main" |
```
OTL, line 8:
```suggestion
echo "deb http://apt.llvm.org/${codename}/ llvm-toolchain-${codename}-${LLVM_VERSION} main" |
```

ci/ubuntu-setup-lld.sh
```diff
@@ -0,0 +1,24 @@
...
+    sudo dd of="/etc/apt/sources.list.d/llvm-toolchain-${codename}-13.list"
```
OTL, line 9: This is the first time to see `dd` is used in such a case. Do you have some examples?

ci/ubuntu-setup-lld.sh
OTL: Is this correct?

ci/ubuntu-setup-lld.sh
OTL: I recommend using `tee` instead of `dd` if `tee` is already installed.

OTL (approved):

PR #586, opened 2022-03-10
Use beta for code coverage

taiki-e:
LLVM source-based code coverage has been stabilized on Rust 1.60 (currently beta).

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -13,7 +13,6 @@ env:
   RUST_BACKTRACE: 1
   RUSTFLAGS: -D warnings
   RUSTUP_MAX_RETRIES: 10
-  nightly: nightly-2021-11-22
 
 jobs:
   coverage:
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -49,7 +48,7 @@ jobs:
           source /opt/ros/foxy/setup.bash
           # for test of arci-ros (roscore, rostopic)
           source /opt/ros/noetic/setup.bash
-          cargo llvm-cov --verbose --all-features --workspace --exclude openrr-internal-codegen --doctests --lcov --output-path lcov.info
+          cargo llvm-cov --verbose --all-features --workspace --exclude openrr-internal-codegen --lcov --output-path lcov.info
```
taiki-e, line 51: This drops doctests because it needs nightly. That said, it is [annoyed by a compiler bug](https://github.com/taiki-e/cargo-llvm-cov/issues/2#issuecomment-1061686546), so dropping it should not have much of an impact.

OTL (approved):

PR #601, opened 2022-05-11
[openrr-planner] Add a test for self collision detection

ssr-yuki:
This PR adds a test of `detect_self` function in CollisionDetector.


OTL (approved):
👍🏼

PR #610, opened 2022-05-27
Add test for sample command

kaaatsu32329:
Add test for sample command parsed by robot_command in openrr-apps.

### DIFF
--- a/openrr-apps/src/bin/robot_command.rs
+++ b/openrr-apps/src/bin/robot_command.rs
@@ -107,6 +107,8 @@ async fn main() -> Result<()> {
 
 #[cfg(test)]
 mod tests {
+    use std::path::Path;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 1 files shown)


openrr-apps/src/bin/robot_command.rs
```diff
@@ -128,4 +128,48 @@ mod tests {
...
     }
+
+    #[test]
+    fn assert_sample_command() {
+        let bin = env!("CARGO_BIN_NAME");
+        let sample_command = RobotCommandArgs::try_parse_from(&[
+            bin,
+            "--config-path=./openrr-apps/config/sample_robot_client_config_for_urdf_viz.toml",
+            "load_commands",
+            "./openrr-apps/command/sample_cmd_urdf_viz.txt",
```
taiki-e, line 139: This does not seem robust as it depends on the current directory.

Instead, I suggest using [the `CARGO_MANIFEST_DIR` environment variable](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) to get the absolute path to the `openrr-apps` directory. For example:

```rs
let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
let sample_command_path = manifest_dir.join("command/sample_cmd_urdf_viz.txt");
```

kaaatsu32329: Thank you for your review. I updated code.

openrr-apps/src/bin/robot_command.rs
```diff
@@ -128,4 +130,56 @@ mod tests {
...
     }
+
+    #[test]
+    fn assert_sample_command() {
+        let bin = env!("CARGO_BIN_NAME");
+        let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+
+        let config_path = manifest_dir.join("config/sample_robot_client_for_urdf_viz.toml");
+        let config_path_arg = "--config-path=".to_string() + config_path.to_str().unwrap();
+        let sample_command_path = manifest_dir.join("command/sample_cmd_urdf_viz.txt");
```
OTL, line 141: "command/sample_cmd_urdf_viz.txt" is repeated in L159.
I think it should be defined as const value.

kaaatsu32329: Thank you for your attention.

kaaatsu32329:
Now, it is only testing load_commands. I also implemented test for other commands. If this style is OK, I will send PR about other commands implementation.

OTL (approved):

PR #624, opened 2022-06-17
Update test for openrr_apps_robot_command

kaaatsu32329:
I implemented test for sample command which is read by openrr_apps_robot_command and update some sample commands.

### DIFF
--- a/openrr-apps/src/bin/robot_command.rs
+++ b/openrr-apps/src/bin/robot_command.rs
@@ -107,7 +107,10 @@ async fn main() -> Result<()> {
 
 #[cfg(test)]
 mod tests {
-    use std::path::Path;
+    use std::{env, path::Path};
+
+    use openrr_command::load_command_file_and_filter;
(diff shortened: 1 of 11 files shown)


openrr-apps/src/bin/robot_command.rs
```diff
@@ -132,6 +135,18 @@ mod tests {
     }
 
     const SAMPLE_COMMAND_PATH: &str = "command/sample_cmd_urdf_viz.txt";
+    const COMMAND_PATHS: [&str; 10] = [
```
OTL, line 138: Is there any reason to divide SAMPLE_COMMAND_PATH and COMMAND_PATHS?
How about pushing SAMPLE_COMMAND_PATH into COMMAND_PATHS if possible?

kaaatsu32329: I think pushing the path is better too. I updated it.

OTL (approved):
Thanks!

Issue #625, opened 2022-06-17
The robot stops immediately if the slider is not touched by velocity_sender.


kaaatsu32329:
It is common in ROS to automatically stop if `/cmd_vel` is not sent for a certain period of time. `velocity_sender` is driven by events in the GUI and it appears that `/cmd_vel` is not sent unless the slider etc. is touched.

PR #628, opened 2022-06-20
Update velocity_sender to fix issue #625

kaaatsu32329:
Update `velocity_sender` to send topic each 100ms.


openrr-gui/src/velocity_sender.rs
```diff
@@ -422,6 +428,11 @@ where
         Command::none()
     }
 
+    fn subscription(&self) -> iced::Subscription<Self::Message> {
+        iced_futures::backend::native::tokio::time::every(std::time::Duration::from_millis(100))
+            .map(messenger)
```
OTL, line 433: I don't know the details, but it seems lambda works fine, right?

```suggestion
            .map(|| Message::Retained)
```

kaaatsu32329: Yes.  I updated it.

OTL:
@taiki-e Do you know a more effective way to implement a timer (periodical behavior) in iced?

openrr-gui/src/velocity_sender.rs
```diff
@@ -411,6 +414,9 @@ where
                     state.update_value(state.value.abs() * sign as f64);
                 }
             }
+            Message::Retained => {
+                self.move_base.send_velocity(&latest_velocity).unwrap();
```
taiki-e, line 418: L422-L424 also sends velocity, so I think you can remove this `send_velocity` and L339's` current_velocity`.

taiki-e:
@OTL AFAIK, using [subscription](https://docs.rs/iced/latest/iced/type.Subscription.html) for time tick is right pattern.

openrr-gui/src/velocity_sender.rs
kaaatsu32329: Thank you for comment. I update as noted above.

taiki-e (approved):

PR #631, opened 2022-06-27
[openrr-planner] Use RobotCollisionDetector in SelfCollisionChecker

ssr-yuki:
This PR is a modified version of #544 , which was reverted in #550 .

The purpose of this PR is introducing a robot model only used for collision checking. The current design shares only one robot model, and the model is used for various applications, e.g. collision checking, planning, solving FK and IK. This is too complicated, and we need to pay attention to what the robot states corresponds to (current pose? target pose?). Introducing a task-focused model can be a solution of this complexity.

Note: I checked that all tests of our private development passed with this change.

### DIFF
--- a/openrr-planner/src/collision/self_collision_checker.rs
+++ b/openrr-planner/src/collision/self_collision_checker.rs
@@ -7,18 +7,20 @@ use serde::{Deserialize, Serialize};
 use tracing::debug;
 
 use crate::{
-    collision::parse_colon_separated_pairs, errors::*, interpolate, CollisionDetector,
-    TrajectoryPoint,
+    collision::{parse_colon_separated_pairs, RobotCollisionDetector},
+    errors::*,
+    interpolate, CollisionDetector, TrajectoryPoint,
 };
 
 pub struct SelfCollisionChecker<N>
 where
     N: RealField + Copy + k::SubsetOf<f64>,
 {
-    pub collision_check_robot: Arc<k::Chain<N>>,
-    pub collision_detector: CollisionDetector<N>,
-    pub collision_pairs: Vec<(String, String)>,
-
+    /// Robot for reference (read only and assumed to hold the latest full states)
+    pub reference_robot: Arc<k::Chain<N>>,
+    /// Robot collision detector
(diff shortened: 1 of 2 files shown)


openrr-planner/src/collision/self_collision_checker.rs
```diff
@@ -7,18 +7,20 @@ use serde::{Deserialize, Serialize};
...
     N: RealField + Copy + k::SubsetOf<f64>,
 {
-    pub collision_check_robot: Arc<k::Chain<N>>,
-    pub collision_detector: CollisionDetector<N>,
-    pub collision_pairs: Vec<(String, String)>,
-
+    /// Robot for reference (read only and assumed to hold the latest full states)
+    pub reference_robot: Arc<k::Chain<N>>,
+    /// Robot collision detector
+    pub robot_collision_detector: RobotCollisionDetector<N>,
```
OTL, line 22: Just comment:

I used to be aggressive, so all members of this structure are public, but I think it is basically better to keep it private if possible.

openrr-planner/src/collision/self_collision_checker.rs
```diff
@@ -175,12 +187,16 @@ fn test_create_self_collision_checker() {
...
             std::time::Duration::new(1, 0),
         )
         .is_err());
 
-    let l_shoulder_yaw_node = robot.find("l_shoulder_yaw").unwrap();
+    let l_shoulder_yaw_node = self_collision_checker
+        .robot_collision_detector
+        .robot
+        .find("l_shoulder_yaw")
+        .unwrap();
```
OTL, line 199: I want to know why you changed the test. Something has been changed by this change? or just improve something?

ssr-yuki: The purposes of the change for this test are:

- to make "sample.urdf" a variable (for improvement)
- to correct collision checking (since the order of the joints is somehow changed...)

OTL: Thanks! I understood that this change causes some shuffling  (it might be making it go the correct direction) about the order of joints comparing the older code.

OTL (approved):

PR #634, opened 2022-07-06
[arci-ros] Rename a client example

### LINKED ISSUE #633: Example name collision
The following warning appears with `cargo test` on the latest main branch (commit 1b4761ebdb6fcbba6a4edd62d79c7d2472426bd1 )

```
warning: output filename collision.
The example target `client` in package `openrr-remote v0.0.6 ([EDITED]/openrr/openrr-remote)` has the same output filename as the example target `client` in package `arci-ros v0.0.6 ([EDITED]/openrr/arci-ros)`.
Colliding filename is: [EDITED]/openrr/target/debug/examples/client
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

We should rename at least one of them.

ssr-yuki:
This PR just closes #633 .


OTL (approved):
thanks!

PR #650, opened 2022-07-22
Add function to edit RobotCommand

### LINKED ISSUE #643: The error occurs when executing a command containing double-cotation marks from `execute_command` in `openrr_apps_robot_commands`
When using double-cotation marks to input dictionary-type arguments that are parsed into Python, they are split by whitespace in the dictionary and passed as separate arguments.

And when a double-cotation mark `"{}"` is passed as an argument, it is entered as an escape character `\"{}\"` and its meaning is changed.

kaaatsu32329:
The purpose of this PR is to fix #643

Edit RobotCommand to remove `\"` and recover ` ` (space) that are required for the argument in the string.

### DIFF
--- a/openrr-command/src/robot_command.rs
+++ b/openrr-command/src/robot_command.rs
@@ -481,6 +481,35 @@ pub fn load_command_file_and_filter(file_path: PathBuf) -> Result<Vec<String>, O
         .collect())
 }
 
+pub fn robot_command_editor(robot_command: &RobotCommand) -> RobotCommand {
+    match robot_command {
+        RobotCommand::ExecuteCommand { command } => {
+            let mut concatenation = false;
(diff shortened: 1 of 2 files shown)


openrr-command/src/robot_command.rs
```diff
@@ -481,6 +481,35 @@ pub fn load_command_file_and_filter(file_path: PathBuf) -> Result<Vec<String>, O
         .collect())
 }
 
+pub fn robot_command_editor(robot_command: &RobotCommand) -> RobotCommand {
```
OTL, line 484: I don't like this function name. It's better if it is a verb for a function. (ex, `filter_escape_doulbe_quote`)

* Is it possible to make it private?
* Because it is handling `ExecuteCommand`, how about doing this around L365?

PR #651, opened 2022-07-28
Add function to edit RobotCommand

kaaatsu32329:
The purpose of this PR is to fix https://github.com/openrr/openrr/issues/643

Edit RobotCommand to remove `\"` and recover `  ` (space) that are required for the argument in the string.

This PR is reopened PR #650

### DIFF
--- a/openrr-command/src/robot_command.rs
+++ b/openrr-command/src/robot_command.rs
@@ -353,6 +353,7 @@ impl RobotCommandExecutor {
                 client.speak(name, &message.join(" "))?.await?;
             }
             RobotCommand::ExecuteCommand { command } => {
+                let command = filter_escape_double_quote(command);
                 let mut iter = command.iter();
                 let cmd_str = iter
                     .next()
@@ -481,6 +482,27 @@ pub fn load_command_file_and_filter(file_path: PathBuf) -> Result<Vec<String>, O
         .collect())
 }
 
+fn filter_escape_double_quote(robot_command: &[String]) -> Vec<String> {
+    let mut is_concatenation = false;
+    let mut command_vector: Vec<String> = Vec::new();
+
+    for element in robot_command {
+        let command_element = element.replace('\"', "");
+        if is_concatenation {
+            let new_command = command_vector
+                .pop()
+                .expect("Parse double quote logic error. Vector command_vector is empty.");
+            command_vector.push(format!("{new_command} {command_element}"));
+        } else {
+            command_vector.push(command_element.to_string());
+        }
+        if element.match_indices('\"').count() == 1 {
+            is_concatenation = !is_concatenation;
+        }
+    }
+    command_vector
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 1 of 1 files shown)


openrr-command/src/robot_command.rs
```diff
@@ -481,6 +482,26 @@ pub fn load_command_file_and_filter(file_path: PathBuf) -> Result<Vec<String>, O
         .collect())
 }
 
+fn filter_escape_double_quote(robot_command: &Vec<String>) -> Vec<String> {
```
OTL, line 485: How about using `&[]` instead of `&Vec<>`?

```suggestion
fn filter_escape_double_quote(robot_command: &[String]) -> Vec<String> {
```

And can you add some unit tests for this function?
Or add the documentation code above, which will be tested automatically.

kaaatsu32329: Thank you. I updated it and added test for the function.

openrr-command/src/robot_command.rs
```diff
@@ -490,4 +511,24 @@ mod test {
         assert_eq!(val.0, 0);
         assert_eq!(val.1, 2);
     }
+
+    #[test]
+    fn test_filter_escape_double_quote() {
+        // This situation assumes that `command_name hoge/fuga "foo bar"` has been entered.
```
OTL, line 517: Please avoid `hoge`, `fuga` which might be Japanese.
And add one more test case with more than three words in the quotes (a more realistic case.)

For example, `command -f "{ type: Message, value: 1.0 }"`

openrr-command/src/robot_command.rs
```diff
@@ -481,6 +482,26 @@ pub fn load_command_file_and_filter(file_path: PathBuf) -> Result<Vec<String>, O
...
 }
 
+fn filter_escape_double_quote(robot_command: &[String]) -> Vec<String> {
+    let mut concatenation = false;
+    let mut command_vector: Vec<String> = Vec::new();
+
+    for element in robot_command {
+        let command_element = element.replace('\"', "");
+        if concatenation {
+            let mut new_command = command_vector.pop().unwrap() + " ";
```
OTL, line 492: It's better to use expect() instead of unwrap().

openrr-command/src/robot_command.rs
```diff
@@ -481,6 +482,26 @@ pub fn load_command_file_and_filter(file_path: PathBuf) -> Result<Vec<String>, O
         .collect())
 }
 
+fn filter_escape_double_quote(robot_command: &[String]) -> Vec<String> {
+    let mut concatenation = false;
```
OTL, line 486: It might be better to use "is_concatenating" to describe what is this meaning.

openrr-command/src/robot_command.rs
OTL, line 492: I think it's more simple like below.

```rust
            let new_command = command_vector.pop().expect("parse double quote logic error");
            command_vector.push(format!("{new_command} {command_element}"));
```

kaaatsu32329: Thank you for comments. I fixed the areas indicated.

OTL (approved):

PR #661, opened 2022-08-05
[openrr-apps] Update sample config for twin arm sample model

kaaatsu32329:
The other day, the sample model `./openrr-planner/sample.urdf` was updated to twin arm robot. But the sample configs in `openrr-apps` have not updated.


OTL (approved):
Oh, thanks!

kaaatsu32329:
For example, openrr_apps_joint_position_sender appears like this.

Screenshot from 2022-08-05 14-50-22

Screenshot from 2022-08-05 14-50-46

PR #669, opened 2022-08-16
Add humble to ROS2 CI

taiki-e:


taiki-e:
it seems r2r doesn't support humble yet

OTL (approved):

PR #676, opened 2022-09-01
[arci-speak] Add test for -audio and -cmd

kaaatsu32329:


arci-speak-cmd/src/lib.rs
```diff
@@ -81,3 +81,18 @@ fn run_local_command(message: &str) -> io::Result<()> {
...
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_local_command() {
+        let local_command = LocalCommand::new();
+
+        let message = "message";
+        let wait = local_command.speak(message);
```
OTL, line 94: If you want to define constant, it's better to use

```rust
const MESSAGE: &str = "message";
```

but you can skip it if you want, like below.

```rust
        let wait = local_command.speak("message");
```

taiki-e (commented):
LGTM aside from a nit that @OTL mentioned.

taiki-e (approved):

PR #704, opened 2022-10-19
Refactor SelfCollisionChecker

ssr-yuki:
This PR aims to refactor SelfCollisionChecker with...

- making error information of self-collision checks detailed
- reduce the number of comparison by reversing an order of `match` and `for`

### DIFF
--- a/openrr-planner/src/collision/self_collision_checker.rs
+++ b/openrr-planner/src/collision/self_collision_checker.rs
@@ -140,9 +140,9 @@ where
             if let Some(names) = self_checker.next() {
                 return Err(Error::Collision {
                     point: match i {
-                        0 => UnfeasibleTrajectory::StartPoint,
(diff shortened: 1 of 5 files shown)


openrr-planner/src/collision/self_collision_checker.rs
```diff
@@ -151,6 +157,11 @@ where
         }
         Ok(())
     }
+
+    // Get the robot model used for collision checking
```
OTL, line 161: Is there any reason why you didn't make this a doc string?

OTL (approved):

openrr-planner/src/collision/self_collision_checker.rs
ssr-yuki: Exactly. But I found I made the same mistake for JointPathPlanner::collision_check_robot. I would like to fix them in another PR.

PR #707, opened 2022-10-20
[openrr-apps] Support JoyGamepad for robot_teleop

taiki-e:
Addresses https://github.com/openrr/openrr-tutorial/pull/1#discussion_r963214012.

TODO:
- ~~needs test~~
- ~~decide the default setting: https://github.com/openrr/openrr/pull/707#discussion_r1000446877~~ #708 
- ~~support custom setting (probably: add JoyGamepadConfig to arci-ros and use it in teleop config)~~

### DIFF
--- a/openrr-apps/config/sample_teleop_config_urdf_viz_joy_gamepad.toml
+++ b/openrr-apps/config/sample_teleop_config_urdf_viz_joy_gamepad.toml
@@ -1,0 +1,38 @@
+robot_config_path = "sample_robot_client_config_for_urdf_viz.toml"
+gamepad = "joy-gamepad"
+
+[joy_gamepad_config]
+axis_map = { '0' = "LeftStickX", '1' = "LeftStickY", '2' = "RightStickX", '3' = "RightStickY", '4' = "DPadX", '5' = "DPadY" }
+button_map = { '0' = "North", '1' = "East", '2' = "South", '3' = "West", '4' = "LeftTrigger", '5' = "RightTrigger", '6' = "LeftTrigger2", '7' = "RightTrigger2" }
+topic_name = "joy"
+
+[control_nodes_config]
+move_base_mode = "base"
+
+[control_nodes_config.joints_pose_sender_config]
+
+[[control_nodes_config.ik_node_teleop_configs]]
+solver_name = "l_arm_ik_solver"
+joint_trajectory_client_name = "l_arm"
+
+[control_nodes_config.ik_node_teleop_configs.config]
+mode = "left i k"
+
+[[control_nodes_config.ik_node_teleop_configs]]
+solver_name = "r_arm_ik_solver"
+joint_trajectory_client_name = "r_arm"
+
+[control_nodes_config.ik_node_teleop_configs.config]
+mode = "right i k"
+
+[[control_nodes_config.joy_joint_teleop_configs]]
+client_name = "l_arm_collision_checked"
+
+[control_nodes_config.joy_joint_teleop_configs.config]
+mode = "l_arm"
+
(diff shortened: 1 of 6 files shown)


openrr-apps/src/bin/robot_teleop.rs
```diff
@@ -155,7 +155,29 @@ async fn main() -> Result<()> {
...
+            let mut axis_map = HashMap::new();
+            axis_map.insert(0, arci::gamepad::Axis::LeftStickX);
+            axis_map.insert(1, arci::gamepad::Axis::LeftStickY);
+            axis_map.insert(2, arci::gamepad::Axis::RightStickX);
+            axis_map.insert(5, arci::gamepad::Axis::RightStickY);
+            axis_map.insert(6, arci::gamepad::Axis::DPadY);
+            axis_map.insert(7, arci::gamepad::Axis::DPadX);
+            switcher
+                .main(arci_ros::JoyGamepad::new("joy", button_map, axis_map))
+                .await;
```
taiki-e, line 176: I copied these default setting from arci-ros's example, but I'm not sure what is the correct default here.

OTL: @kaaatsu32329 Could you try this?
Anyway, ROS joy_node can be configurated from ROS parameter system. It might not be so much problem if the button/axis bind is not comfortable. (But we recommend PS4/PS5 joy in openrr, it should be for PS4 as default)

openrr-apps/src/robot_teleop_config.rs
```diff
@@ -56,6 +59,9 @@ pub struct RobotTeleopConfig {
     pub control_nodes_config: ControlNodesConfig,
     #[serde(default)]
     pub gil_gamepad_config: GilGamepadConfig,
+    #[cfg(feature = "ros")]
+    #[serde(default)]
+    pub joy_gamepad_config: JoyGamepadConfig,
```
OTL, line 64: I feel that it's better to use `RosJoyGamepadConfig`.
Someone may feel that `joy` means ROS actually, but it's not so usual.

openrr-apps/src/bin/robot_teleop.rs
```diff
@@ -155,7 +155,20 @@ async fn main() -> Result<()> {
         }
         #[cfg(windows)]
         GamepadKind::Builtin(BuiltinGamepad::Keyboard) => {
-            tracing::warn!("`gamepad = \"Keyboard\"` is not supported on windows");
+            anyhow::bail!("`gamepad = \"keyboard\"` is not supported on windows");
+        }
+        #[cfg(feature = "ros")]
+        GamepadKind::Builtin(BuiltinGamepad::JoyGamepad) => {
+            arci_ros::init("joy_gamepad");
```
OTL, line 162: See #720

openrr-apps/src/robot_teleop_config.rs
kaaatsu32329: #721

kaaatsu32329: @OTL Should I rename the file as well?

openrr-apps/config/sample_teleop_config_urdf_viz_joy_gamepad.toml
```diff
@@ -0,0 +1,38 @@
+robot_config_path = "sample_robot_client_config_for_urdf_viz.toml"
```
OTL, line 1: sample_teleop_config_urdf_viz_ros_joy_gamepad.toml might be better, but it's OK.

OTL (approved):

PR #708, opened 2022-10-21
[openrr-apps] Fix JoyGamepad layout for openrr_apps_robot_teleop

kaaatsu32329:
I test this and add arci_ros::init, then change some layout.


taiki-e (approved):
Thanks!

openrr-apps/src/bin/robot_teleop.rs
```diff
@@ -159,18 +159,23 @@ async fn main() -> Result<()> {
         }
         #[cfg(feature = "ros")]
         GamepadKind::Builtin(BuiltinGamepad::JoyGamepad) => {
+            arci_ros::init("joy_gamepad");
```
OTL, line 162: @kaaatsu32329 @taiki-e 
Calling arci_ros::init() (= rosrust::init()) is not good idea.

PR #719, opened 2022-10-28
Revert "Update clap requirement from 3.1 to 4.0"

taiki-e:
Reverts openrr/openrr#692


OTL (approved):
Could you tell me what happened when it was upgraded, if you know?

PR #720, opened 2022-10-31
[openrr-apps] Fix to avoid double initialization

kaaatsu32329:


taiki-e (approved):

OTL:
I'm sorry but this don't fix the problem actually.

OTL:
Ah, it's my mistake. This may work fine!
I think it's better to change the node name.

openrr-apps/src/bin/robot_teleop.rs
```diff
@@ -159,7 +159,9 @@ async fn main() -> Result<()> {
         }
         #[cfg(feature = "ros")]
         GamepadKind::Builtin(BuiltinGamepad::JoyGamepad) => {
-            arci_ros::init("joy_gamepad");
+            if !use_ros {
+                arci_ros::init("joy_gamepad");
```
OTL, line 163: Could you change the node name from "joy_gamepad" to `env!("CARGO_BIN_NAME")`?

OTL (approved):

PR #733, opened 2022-11-26
Update k requirement from 0.28 to 0.29

ssr-yuki:
Some tests are fixed because updating `k` changed the order of joints.


OTL (approved):
Geat fix!!

PR #737, opened 2022-12-07
Use cSpell for spell check

taiki-e:


.github/.cspell/dictionary.txt
```diff
@@ -144,33 +130,29 @@ readline
 README
 recv
 Reloadable
+repr
+repr
```
OTL, line 134: Could you try uniq command to detect other possible duplications?

OTL (commented):
I think it's OK but it will lose GitHub integration, right?

taiki-e:
Yes, but this allows us to share the configuration file with the vs code extension, which I consider more useful than GHA action that sometimes breaks with odd errors.

OTL (approved):
Thanks!

PR #744, opened 2022-12-09
Improve spell-check

taiki-e:
- Collect rust dependency names automatically.
- Address reviews from @OTL that done in other repos.
- Sync cspell config with other repos.


tools/spell-check.sh
```diff
@@ -0,0 +1,45 @@
...
+#    ./tools/spell-check.sh
+
+metadata=$(cargo metadata --format-version=1 --all-features --no-deps)
+dependencies=''
+for id in $(jq <<<"${metadata}" '.workspace_members[]'); do
+    dependencies+=$'\n'
+    dependencies+=$(jq <<<"${metadata}" ".packages[] | select(.id == ${id})" | jq -r '.dependencies[].name')
+done
+cat >.github/.cspell/rust-dependencies.txt <<EOF
+// This file is @generated by $(basename "$0").
```
OTL, line 17: I want to confirm that you intentionally used @ in front of generated. What does it mean?

taiki-e: It is an annotation indicating that the file was generated by the program; Cargo.lock has it as well, and some tools (formatters, license header checkers, etc.) can exclude those files.

```console
$ cat Cargo.lock | head -2
# This file is automatically @generated by Cargo.
# It is not intended for manual editing.
```

OTL: I see. Thank you for the clear explanation!!!

OTL (approved):

PR #762, opened 2023-01-17
[arci-speak-audio] Disable default-features of rodio

taiki-e:
Addresses an advisory from transitively dep of rodio.


OTL (approved):
LGTM!

PR #764, opened 2023-01-23
Use workspace inheritance for [package] table

taiki-e:
Use workspace inheritance stabilized in Rust 1.64: https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds

Refs: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table


OTL (approved):
👍🏼👍🏼👍🏼👍🏼

PR #775, opened 2023-02-07
[arci-ros] Fix attribute of some tests for arci-ros

kaaatsu32329:
#767

### DIFF
--- a/arci-ros/tests/test_localization_client.rs
+++ b/arci-ros/tests/test_localization_client.rs
@@ -1,4 +1,5 @@
-#[cfg(target_os = "linux")]
+#![cfg(target_os = "linux")]
+
 mod msg {
     rosrust::rosmsg_include!(
         geometry_msgs / PoseWithCovarianceStamped,
--- a/arci-ros/tests/test_robot_client.rs
+++ b/arci-ros/tests/test_robot_client.rs
@@ -1,4 +1,5 @@
-#[cfg(target_os = "linux")]
+#![cfg(target_os = "linux")]
+
 use arci::{JointTrajectoryClient, TrajectoryPoint};
 mod msg {
     rosrust::rosmsg_include!(sensor_msgs / JointState, trajectory_msgs / JointTrajectory);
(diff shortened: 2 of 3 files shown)


OTL:
I wrote that it can be run on Mac, but it's not true.
I think it needs roscore which is not included in rosrust.

OTL (changes requested):
So sorry, but revert to linux only.

kaaatsu32329:
The cause of the test error is probably in the `test_speak_client`. Maybe duration time is not enough.

https://github.com/openrr/openrr/blob/027651333dc4beeadb1e34b18a3a136262d987c1/arci-ros/tests/test_speak_client.rs#L27

But I have no idea to implement like below in `RosEspeakClient`

```rust
while rosrust::is_ok() && foo_publisher.subscriber_count() == 0 {
...
}
```

`RosEspeakClient` has no public publisher, only a private one.

https://github.com/openrr/openrr/blob/027651333dc4beeadb1e34b18a3a136262d987c1/arci-ros/src/ros_speak_client.rs#L10

So I changed the duration time to longer.

kaaatsu32329:
In my local pc, sometimes occurred errors like the one below.

```bash
~/openrr$ cargo test --package arci-ros --test test_speak_client -- test_speak_client --exact --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.17s
    Running tests/test_speak_client.rs (target/debug/deps/test_speak_client-b605abb8344002dd)
running 1 test
arci ros speak client test
Running roscore on port: 17894
[Speaker]: "Good morning!"
test test_speak_client has been running for over 60 seconds
```

This is an error in GitHub Actions.
https://github.com/openrr/openrr/actions/runs/4102118427/jobs/7095068451
>      Running tests/test_speak_client.rs (/__w/openrr/openrr/target/debug/deps/test_speak_client-203d71f4294fff1f)
> 
> running 1 test
> test test_speak_client has been running for over 60 seconds

I think they are in the same situation.

OTL (approved):

PR #779, opened 2023-02-08
Add batch to README

kaaatsu32329:
Like below. Add batch to link of `OpenRR Turorial`.

Screenshot from 2023-02-08 13-15-15


OTL (approved):
Thanks!

PR #794, opened 2023-03-10
[arci] Add `MotorDrive` traits

kaaatsu32329:


arci/src/traits/motor_drive.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn get_motor_velocity(&self) -> Result<Option<ActuatorState>, Error>;
+}
+
+#[auto_impl(Box, Arc)]
+pub trait MotorDriveEffort: Send + Sync {
+    fn set_motor_effort(&self, actuator_state: ActuatorState) -> Result<(), Error>;
+    fn get_motor_effort(&self) -> Result<Option<ActuatorState>, Error>;
+}
+
+pub enum ActuatorState {
```
OTL, line 23: We should not share this with among position/velocity/torque.
It's better to use different types to avoid some bugs.
I feel use f64 directly is enough...

kaaatsu32329:
I will add tests for new `arci` traits after the change is merged.

arci/src/clients/dummy_motor_drive.rs
OTL, line 1: Just a comment:
I think it's better to use Mock system, for example (`mockall` crate).
When I started implementing dummy_*.rs, I did not know about it.

OTL (approved):
LGTM

PR #812, opened 2023-04-03
Enable region coverage on codecov

taiki-e:
https://github.com/taiki-e/cargo-llvm-cov/releases/tag/v0.5.12

Catch cases where some regions are covered and others are not covered in the same line. For example, if the case that returns Ok is tested, but the case that returns Err is not, the line is marked as partially covered.

~~Blocked on https://github.com/taiki-e/cargo-llvm-cov/issues/252.~~


OTL (approved):
Thanks！

Issue #814, opened 2023-04-12
Inactive links to tutorial books


abrzozowski:
The links to the tutorial book are inactive: https://github.com/openrr/openrr/blob/97c83ca9463d84baf223ae856ce3125907746ab0/README.md?plain=1#L113

Have those links changed?

OTL:
Thank you for the information!!! #816 will solve it.

PR #815, opened 2023-04-12
Fix inactive links to tutorial books

abrzozowski:
Solve the issue reported in https://github.com/openrr/openrr/issues/814


OTL (approved):
Thanks!

PR #817, opened 2023-04-20
Remove ROS1 melodic from CI

taiki-e:
This will be EoL at the end of this month.
https://wiki.ros.org/Distributions


OTL:
We can remove it now.

OTL (approved):

PR #819, opened 2023-04-27
[openrr-base] Add controller for diffrential drive

kaaatsu32329:


openrr-base/src/utils.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+impl Default for BaseVelocityTimestamped {
+    fn default() -> Self {
+        Self {
+            base_velocity: BaseVelocity::default(),
+            timestamp: std::time::Instant::now(),
+        }
+    }
+}
```
OTL, line 16: I think it's nice to add a function to create this with velocity and Instant::now();

```rust
fn new_with_now(vel :BaseVelocity) -> Self {
}
```

openrr-base/src/base_odometry.rs
```diff
@@ -0,0 +1,66 @@
...
+        let theta = pose.to_owned().rotation.angle();
+        let delta = update_time.to_owned().elapsed().as_secs_f64();
+
+        *pose = Isometry2::new(
+            Vector2::new(
+                x + (velocity.x * theta.cos() - velocity.y * theta.sin()) * delta,
+                y + (velocity.x * theta.sin() + velocity.y * theta.cos()) * delta,
+            ),
+            theta + velocity.theta * delta,
+        );
```
OTL, line 33: I feel better to use Isometry2::mut.
I mean define delta_pose as Isometry2, then you can do

```rust
*pose *= delta_pose;
```

(I don't know about lock, sorry.)

openrr-base/src/base/differential_drive.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.controller
+            .right
+            .set_motor_velocity(wheels_vel[1])
+            .unwrap();
+
+        let wheels_fdb = vec![
+            self.controller.left.get_motor_velocity().unwrap(),
+            self.controller.right.get_motor_velocity().unwrap(),
+        ];
+        let fdb_vel = self.transform_velocity_wheel_to_base(&wheels_fdb).unwrap();
```
OTL, line 97: I can't understand `fdb` means. I recommend not to use abbreviation.

OTL (commented):
We might need some more review, but please update once.

openrr-base/src/odometry.rs
```diff
@@ -0,0 +1,109 @@
+use arci::{BaseVelocity, Isometry2, Vector2};
+use parking_lot::Mutex;
+
+use crate::Error;
+
+const ODOMETRY_TIMEOUT_MILLIS: u128 = 5000;
```
OTL, line 6: This should be a variable, and it should be possible to be disabled.

openrr-base/src/odometry.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[derive(Debug)]
+pub struct Odometry {
+    /// If `None`, it is odometry lost.
+    position: Mutex<Option<Isometry2<f64>>>,
+    last_update_timestamp: Mutex<Option<std::time::Instant>>,
+}
+
+impl Odometry {
+    pub fn new_from_position_with_now(position: Isometry2<f64>) -> Self {
```
OTL, line 16: How about `from_pose()`?
position sounds like it has no rotation.

openrr-base/src/odometry.rs
```diff
@@ -0,0 +1,109 @@
...
+        Self {
+            position: Mutex::new(Some(position)),
+            last_update_timestamp: Mutex::new(None),
+        }
+    }
+
+    pub fn translate(&self, velocity: &BaseVelocity) -> Result<(), Error> {
```
OTL, line 23: translate means move only position. This contains rotation.
So it's better to avoid the name.
My suggestion is `update_by_velocity()`, `transform()`.

openrr-base/src/odometry.rs
OTL: I think it's OK to keep the value if the last velocity is small enough.

openrr-base/src/odometry.rs
```diff
@@ -0,0 +1,109 @@
...
+                    *position = None;
+                    *locked_update_time = Some(std::time::Instant::now());
+                    Err(Error::CurrentPositionUnknown)
+                } else {
+                    let x = p.translation.x;
+                    let y = p.translation.y;
+                    let theta = p.rotation.angle();
+                    let dt = last_update_time.elapsed().as_secs_f64();
+
+                    *position = Some(Isometry2::new(
```
OTL, line 44: let's use mul!

openrr-base/src/robot_velocity_status.rs
```diff
@@ -0,0 +1,160 @@
+use arci::BaseVelocity;
+use parking_lot::Mutex;
+
+use crate::*;
+
+#[derive(Debug, Default)]
+pub struct RobotVelocityStatus {
+    velocity: Mutex<BaseVelocity>,
+    feedback_velocity: Mutex<BaseVelocity>,
```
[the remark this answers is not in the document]
kaaatsu32329: @OTL 
It is intended to be a variable that holds the speed of the robot calculated from the feedback from the motor.

openrr-base/src/robot_velocity_status.rs
```diff
@@ -0,0 +1,160 @@
...
+            feedback_velocity: Mutex::new(BaseVelocity::default()),
+            max_velocity,
+            min_velocity,
+            max_acceleration,
+            min_acceleration,
+            last_sent_log: Mutex::new(BaseVelocityTimestamped::default()),
+        }
+    }
+
+    pub fn get_velocity(&self) -> BaseVelocity {
```
OTL, line 44: I feel better to remove `get_`

```suggestion
    pub fn velocity(&self) -> BaseVelocity {
```

openrr-base/src/robot_velocity_status.rs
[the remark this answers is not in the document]
OTL: I can understand now! We use `current_` or `_stete` for that meaning in other code of this project.
How about using `velocity_state`?

(BTW, I fell `current_` is not so good name for now)

OTL (approved):
LGTM！

PR #834, opened 2023-06-15
Improve codegen

taiki-e:
Remove the need for these two:

https://github.com/openrr/openrr/blob/4c2fcb8e649dab5452889c32d56f4f286af33aec/arci/README.md#L11-L13


OTL (approved):
👍🏼

PR #846, opened 2023-07-31
Remove max size for GUI

kaaatsu32329:
In current GUI, we cannot access slide bar in the bottom

Screenshot 2023-08-01 at 1 40 01


OTL (approved):
Good catch!

taiki-e (approved):

taiki-e:
CI failure should be fixed by rebasing.

PR #851, opened 2023-08-29
Improve configs for collision checking

### LINKED ISSUE #251: Update openrr-apps sample config
Current config does not work for collision checking (The checked pair is only one and useless).

Use more pairs like below.

https://github.com/OTL/openrr-apps-config-sample/blob/main/config/robot_config.toml

H1rono:
This PR closes #251.


OTL (approved):
Thanks! Great!

PR #852, opened 2023-08-31
Add tests of `LaserScan2D`

H1rono:


arci/src/clients/dummy_laser_scan.rs
```diff
@@ -0,0 +1,27 @@
+use crate::{Error, LaserScan2D, Scan2D};
+
+#[derive(Debug, Clone, Default)]
+pub struct DummyLaserScan2D;
```
taiki-e, line 4: Like other Dummy* types, we may want to store the value and allow it to be changed.
https://github.com/openrr/openrr/blob/8489d2c47339bcf9ee2879f1bc315a64489587fa/arci/src/clients/dummy_motor_drive.rs#L9-L11

OTL: @H1rono Do you have any update about this?

H1rono: Not yet. What kind of value we may want to change?

OTL: Scan2D data which is returned by `fn current_scan()`.
(It might make some future tests easy)

taiki-e (approved):
Thanks!

Issue #859, opened 2023-09-11
arci-ros2: Implement more types


OTL:
- [x] navigation (buggy #858)
- [x] move_base
- [x] joint_trajectory_client (ros2_control)
- [ ] localization
- [ ] transform_resolver
- [ ] laser_scan
- [ ] motor_drive
- [ ] gamepad
- [ ] speaker (We don't need this or we don't have good implementation for this in ROS2 world)

PR #860, opened 2023-09-13
Add `Ros2LocalizationClient`

kaaatsu32329:
#859 Localization

### DIFF
--- a/arci-ros2/src/lib.rs
+++ b/arci-ros2/src/lib.rs
@@ -17,6 +17,7 @@ mod node;
 mod plugin;
 mod ros2_control;
 mod ros2_laser_scan;
+mod ros2_localization_client;
 mod utils;
 
 pub use cmd_vel_move_base::*;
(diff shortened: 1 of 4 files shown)


taiki-e (changes requested):
- Could you update plugin.rs? (like [this](https://github.com/openrr/openrr/commit/196d7861973d4b9d42cc7c2d10a563f45d7a2ba7#diff-353541611e293847f166fd8d0225cd6932b6e707d1a35812bacd2944312e343d))
- Could you add a test for request_final_nomotion_update_hack=true case?
- ~~Could you add a comment on why request_final_nomotion_update_hack is considered a "hack" and why that is needed?~~

taiki-e:
LGTM, but you need to resolve conflicts.

OTL:
@taiki-e Please hijack this.

taiki-e (approved):

PR #868, opened 2023-09-20
[arci-ros2] Use Node instead of r2r::Context in API

taiki-e:
This also unifies the documentation style.

Split from #863.


OTL (approved):
GJ!

Issue #870, opened 2023-09-21
arci-ros2: fix plugins


taiki-e:
AFAIK, it is currently broken.

PR #872, opened 2023-09-22
[openrr-plugin] Handle futures efficiently

taiki-e:
Part of #870 (fix it by combining this with #863)

- Removes async -> blocking -> async translation.
- Improve codegen to support JointTrajectoryClient/Gamepad.


OTL (approved):
Sounds great!

PR #886, opened 2023-11-17
Use [lints] in Cargo.toml and apply more lints

taiki-e:
- Use [lints] in Cargo.toml
  https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo

- Apply some lints that are currently applied only in a few crates to all crates.


OTL (approved):
👍🏼

PR #921, opened 2024-07-31
Use std::sync::{LazyLock,OnceLock} instead of once_cell

taiki-e:
std::sync::LazyLock has been stabilized in Rust 1.80.

The first to third commits are CI fixes from #920.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,6 @@ nalgebra = "0.30"
 ncollide3d = "0.33"
 nix = { version = "0.29", features = ["signal"] }
 num-traits = "0.2"
-once_cell = "1"
 paste = "1"
 portpicker = "0.1"
 prettyplease = "=0.2.20"
--- a/arci-ros/Cargo.toml
+++ b/arci-ros/Cargo.toml
@@ -14,7 +14,6 @@ arci.workspace = true
 auto_impl.workspace = true
 flume.workspace = true
 nalgebra.workspace = true
-once_cell.workspace = true
 paste.workspace = true
 ros-nalgebra.workspace = true
 rosrust.workspace = true
@@ -37,7 +36,6 @@ openrr-teleop.workspace = true
 
 [target.'cfg(unix)'.dev-dependencies]
 nix.workspace = true
-once_cell.workspace = true
 portpicker.workspace = true
 
 [lints]
--- a/arci-ros/src/ros_control/ros_control_action_client.rs
+++ b/arci-ros/src/ros_control/ros_control_action_client.rs
@@ -1,5 +1,5 @@
 use std::{
-    sync::{Arc, Mutex},
+    sync::{Arc, LazyLock, Mutex},
     time::Duration,
 };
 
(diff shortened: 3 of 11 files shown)


OTL (approved):
👍🏼

mtb0x1:
@taiki-e  This commit is a change breaker, the project will not compile with any older Rust version from now on. 
Not everyone is on 1.80.0, we need to specify the compatible rust version so users can be aware of it, this can be done https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field.

In Cargo.toml : 
```toml
[package]
#...
rust-version = "1.80.0"
```

taiki-e:
- There have been discussions for years that increasing MSRV is a non-breaking change, and I believe there is a broad consensus within the ecosystem that it is. See also discussions I mentioned in [this comment](https://github.com/eupn/macrotest/issues/68#issuecomment-1001239647).
- If we add that field, we should check it in the CI to make sure it is not outdated.
  - However, TBH, I don't think it is realistic to support MSRV in this style decently on a project of this scale, because our dependencies increase MSRV frequently.
    - In this PR's case, we just happened to increase our MSRV earlier than our dependencies.
  - The only realistic thing I can think of is the “latest Rust version only (MSRV:1)" approach that rust-lang/cargo does through automation. https://github.com/rust-lang/cargo/pull/14303
    - lockfile is an another possible approach but not very good for library users as I said in [this comment](https://github.com/smol-rs/async-io/issues/93#issuecomment-1255793764).
- I think the inferred MSRV if the field is not set is “the version actually tested by CI”. We have never in the past tested or actually supported anything other than the (latest)stable/beta/nightly by CI.

The actions I think we can actually take here are:

- Update the documentation to clarify that only the latest stable is supported. Like [roaring-rs does](https://github.com/RoaringBitmap/roaring-rs#rust-version-policy).
- Update the documentation to recommend specifying git revision when using openrr from git.
- Update the documentation and recommend users to commit the lockfile. Like [cargo docs](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control).
- Increase the frequency of releases so that most users do not have to use openrr via git.
- Explicit “only latest stable” support using rust-lang/cargo style rust-version. (This will reject compiling with an older version even if it is actually possible to compile with an older version.)

mtb0x1:
> * I think the inferred MSRV if the field is not set is “the version actually tested by CI”. We have never in the past tested or actually supported anything other than the (latest)stable / beta / nightly by CI.

I didn't know about this, and since it's the case this totally fine by me. you might need to check the ci though I see nightly version is fixed to a specific version (https://github.com/openrr/openrr/blob/f2e81a65a41ca5b5b72549e8c89a6c8b29c233e0/.github/workflows/ci.yml#L19C3-L19C10).

As for the suggested actions, the first one would be nice. Something in the docs like "we support latest stable version" is enough and it should infer most the rest. Releasing crates more often can be top but I am afraid this will bring extra "maintenance" to the team (at least docs need to pass the checks) so it's up to you :).

In any case, thank you for your detailed response.

PR #930, opened 2024-09-04
Use basedir when loading from urdf file

soraxas:
Use the `_with_base_dir` variant when loading from the file, otherwise, it will receive error like
```sh
2024-09-04T01:49:24.666406Z ERROR openrr_planner::collision::urdf: ../meshes/collision/link0.stl not found
2024-09-04T01:49:24.666437Z ERROR openrr_planner::collision::urdf: ../meshes/collision/link1.stl not found
2024-09-04T01:49:24.666442Z ERROR openrr_planner::collision::urdf: ../meshes/collision/link2.stl not found
2024-09-04T01:49:24.666447Z ERROR openrr_planner::collision::urdf: ../meshes/collision/link3.stl not found
2024-09-04T01:49:24.666451Z ERROR openrr_planner::collision::urdf: ../meshes/collision/link4.stl not found
2024-09-04T01:49:24.666456Z ERROR openrr_planner::collision::urdf: ../meshes/collision/link5.stl not found
2024-09-04T01:49:24.666461Z ERROR openrr_planner::collision::urdf: ../meshes/collision/link6.stl not found
```
if the urdf has relative path

### DIFF
--- a/openrr-planner/src/collision/collision_detector.rs
+++ b/openrr-planner/src/collision/collision_detector.rs
@@ -452,13 +452,21 @@ mod tests {
 
 /// Convert urdf object into openrr_planner/ncollide3d object
 pub trait FromUrdf {
-    fn from_urdf_robot(robot: &urdf_rs::Robot) -> Self;
-    fn from_urdf_file<P>(path: P) -> ::std::result::Result<Self, urdf_rs::UrdfError>
+    fn from_urdf_robot_with_base_dir(robot: &urdf_rs::Robot, path: Option<&Path>) -> Self;
+    fn from_urdf_robot(robot: &urdf_rs::Robot) -> Self
+    where
+        Self: std::marker::Sized,
+    {
+        Self::from_urdf_robot_with_base_dir(robot, None::<&Path>)
+    }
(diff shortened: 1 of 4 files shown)


openrr-planner/src/collision/collision_detector.rs
```diff
@@ -452,13 +452,21 @@ mod tests {
 
 /// Convert urdf object into openrr_planner/ncollide3d object
 pub trait FromUrdf {
-    fn from_urdf_robot(robot: &urdf_rs::Robot) -> Self;
-    fn from_urdf_file<P>(path: P) -> ::std::result::Result<Self, urdf_rs::UrdfError>
+    fn from_urdf_robot_with_base_dir(robot: &urdf_rs::Robot, path: Option<&Path>) -> Self;
+    fn from_urdf_robot(robot: &urdf_rs::Robot) -> Self
```
taiki-e, line 456: If there is no longer any code that actually calls this API, in other words, if we should always call from_urdf_robot_with_base_dir, we may want to change from_urdf_robot to take `base_path: Option<&Path>` instead of adding from_urdf_robot_with_base_dir.

soraxas: It's following the same convention as the other existing `xxx_with_base_dir`

taiki-e (approved):
LGTM, thanks!

PR #941, opened 2024-11-05
[arci-gamepad-gilrs] Update send command process

kaaatsu32329:
Add new method to send command avoid using event handler.

### DIFF
--- a/arci-gamepad-gilrs/src/lib.rs
+++ b/arci-gamepad-gilrs/src/lib.rs
@@ -346,7 +346,7 @@ impl GilGamepad {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone, JsonSchema)]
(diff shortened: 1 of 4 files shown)


arci-gamepad-gilrs/src/lib.rs
```diff
@@ -278,17 +278,95 @@ impl GilGamepad {
     pub fn new_from_config(config: GilGamepadConfig) -> Self {
         Self::new(config.device_id, config.map)
     }
+
+    /// This is another implementation for send command all the times
```
OTL, line 282: Is it difficult to reuse the code of `new()`? I feel that you copy the code too much.

kaaatsu32329: I think so too and tried to reduce code on below commit.

OTL (approved):
LGTM!

PR #950, opened 2024-12-24
Document how to publish

taiki-e:
Similar to https://github.com/openrr/urdf-viz/pull/275, but having troubleshooting note for when do not complete publishing all crates.


OTL (approved):
Thanks

PR #953, opened 2025-01-09
Implement current_velocity for arci_ros2

kaaatsu32329:


taiki-e:
Could you add a test for this? (Something like [this](https://github.com/openrr/openrr/blob/862788b576e779bf1c88c284c642d3d7beb57e3b/arci-ros2/tests/test_ros2_laser_scan.rs#L22))

taiki-e (approved):
Thanks!

